PR #5, opened 2016-07-22
3.x bitfinex support

BarnumD:
Add bitfinex support for 3.x


carlos8f:
wow 3.x hasn't even landed and already a PR 🎉

PR #21, opened 2016-08-16
WIP kraken support

grigio:


grigio:
Minimal support #9 
- it calls the public api to get BTCUSD rates

I'm not sure if Kraken supports &limit parameter

grigio:
@carlos8f I see teh data processing, but I don't see the graph on localhost.. I get "no data"

carlos8f:
hmm, I'll test it out

grigio:
The backfill time is always the same

carlos8f:
fixed it in master, the `since` param is actually microtime, not a standard timestamp

grigio:
Thanks!

PR #23, opened 2016-08-16
Fixed exchanges.json and reverted some of my suggestions... Double Check!

ghostnegotiator:
Double check it before merging... :panda_face:


ghostnegotiator:
Accidentially flipped currency & asset will be fixed by tonight and adding more currencies including USDT.

ghostnegotiator:
Fixed... these are all currencies & most errors removed...

_This is not tested_

carlos8f:
thanks for the PR! one thing though, in zenbot, FYI, the trade quote price should always be in the currency and the size is in the asset. I only tested BTC/USDT, but I had to reverse the pair in the id to `USDT_BTC` for the API calls to work. it seems `USDT` is a currency in their system, and BTC is the asset (which makes sense).

ghostnegotiator:
Hopefly I won't mess up next time, apparently I lost focus on it
eventually. Although its a manual batch job without IDE where errors occur
easily. I definitely need to install an IDE asap.

PR #26, opened 2016-08-17
More kraken currencies + script

grigio:


grigio:
#9 Now the missing part is the buy/sell via API

carlos8f:
nice 👍

PR #34, opened 2016-08-22
Added All Poloniex USDT Pairs

JFD3D:
I've added all the USDT pairs for Poloniex. I'm working on getting Poloniex live trading implemented but it still needs a lot of testing since there is no 'market' price option in the Poloniex API and while I've been able to connect and place orders I am running into trouble getting the orders to actually fill instead of just placing an order and hoping it gets filled. If the buy/sell query doesn't completely fill orders and just sits there it will cause zenbot to only see the remaining balance (5% of total balance with zenbot's 95% trade defaults or whatever balance is remaining). I will probably end up having to tweak some things and have the trade logic look at the orderbooks and modify sitting orders.

I'm currently testing the 'fillOrKill', 'immediateOrCancel', and 'postOnly' parameters to figure out the most reliable method. I am also running into some errors with the nonce value (probably due to async) that Poloniex API requires but I should be able to figure that out.

Any suggestions would be much appreciated :)

### DIFF
--- a/plugins/poloniex/exchange.json
+++ b/plugins/poloniex/exchange.json
@@ -9,7 +9,7 @@
       "min_size":0.01,
       "max_size":10000,
       "increment":0.01,
-      "label":"BTC/USDT"
+      "label":"USDT/BTC"
     },
     {
       "id":"USDT_DASH",
--- a/plugins/poloniex/exchange.json
+++ b/plugins/poloniex/exchange.json
@@ -3,6 +3,87 @@
   "rest_url": "https://poloniex.com/public",
   "products": [
     {
+      "id":"USDT_BTC",
+      "asset":"BTC",
+      "currency":"USDT",
+      "min_size":0.01,
+      "max_size":10000,
+      "increment":0.01,
+      "label":"BTC/USDT"
+    },
+    {
+      "id":"USDT_DASH",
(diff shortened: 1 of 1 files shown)


carlos8f:
very interesting, I didn't know Poloniex lacked the market option. It would certainly come in handy to have a trade strategy (alternate logic file) for using the order book instead of market flag. Hopefully we can convert it later into an abstraction for order books, so will be portable to other exchanges.

carlos8f:
worth noting that i'm currently doing some refactoring to the default_logic.js for 3.5.0. the ROI went down to 1.1 at some point (when i added the fee calc probably) so i've been optimizing it back into 1.8 territory today

xangma:
If the order book were to be used the balance/volume to be traded would need to be checked to make sure is below the amount of volume for that order price; if it was over, a couple of prices may have to be bought though (and total trade cost would need to take that into account), if that makes any sense?

carlos8f:
@xangma yes, the market depth ideally would be tracked to calculate slippage

current default_logic.js just simulates the trade price (then if using the "run" command, syncs the real balance later and recalculates the ROI). this is handy for simulations but if you're trading large volumes on a real exchange you'd need more accurate price quotes.

PR #55, opened 2016-08-26
adding docker support

egorbenko:


carlos8f:
awesome :) will look at this a little later

grigio:
I suggest to add a volume also for `mongo` so you won't lose the data if you restart zenbot, example ./data/ and put it in gitignore

egorbenko:
@grigio good point, added separate container for mongodb-data. 
p.s. you won't lose data as long as you don't remove mongodb container
- moved Dockerfile to main directory
- added environment check for docker
- added more comments

grigio:
@egorbenko cool, I hope it will be merged, Is there a reason to check be in the in docker?

To me this is enough

``` javascript
c.mongo_url = "mongodb://" + (process.env.MONGODB_PORT_27017_TCP_ADDR || "localhost")+ ":27017/zenbrain"
```

instead of 

https://github.com/egorbenko/zenbot/blob/6a7e34124db0480fd539540b5ec3c2b8f9a91cbb/config_sample.js#L6-L9

egorbenko:
@grigio that works too :), thank you for the comments

BarnumD:
FYI, on my docker image i've made some launch scripts so that the user can decide exactly which components to run and what parameters to provide.  I'm not sure if you think it will be valuable here:

https://github.com/BarnumD/docker-zenbot

egorbenko:
@BarnumD you can already do that with docker-compose. 
`docker-compose up mongodb reducer btc-usd server`

grigio:
Personally I like more `docker-compose` because it is a more standard way to deal with multiple processes in docker.

Another way with less overhead is with GNU parallel, a single command for multiple and connected processes

```
❯ parallel --ungroup ::: "ping google.com" "ping github.com"                                               ⏎
PING google.com (172.217.16.14): 56 data bytes
64 bytes from 172.217.16.14: icmp_seq=0 ttl=56 time=18.626 ms
PING github.com (192.30.253.113): 56 data bytes
64 bytes from 192.30.253.113: icmp_seq=0 ttl=53 time=111.576 ms
64 bytes from 172.217.16.14: icmp_seq=1 ttl=56 time=20.574 ms
64 bytes from 192.30.253.113: icmp_seq=1 ttl=53 time=111.746 ms
64 bytes from 172.217.16.14: icmp_seq=2 ttl=56 time=20.605 ms
64 bytes from 192.30.253.113: icmp_seq=2 ttl=53 time=111.267 ms
```

I didn't test it inside a docker container but it should work.

BarnumD:
You should pin node to at least a major version.

PR #94, opened 2017-01-07
Easier to get started with Kraken trading

tabenius:
I committed these changes to a fork after I got zenbot to work on Kraken by generalizing code I found in discussions at https://github.com/carlos8f/zenbot/issues/9


tabenius:
The code for the balance is suboptimal, the ROI will be wrong if the kraken balance comes after the REAL TRADING code is triggered for the first time. I have quite a lot of code to read.. some pointers would be very helpful to I can help fix it.

tabenius:
broken, removing the request.

PR #105, opened 2017-01-22
Include network ip in sim result link

s-ftf:
Get network IP for sim result graph url in case server is running on networked machine instead of localhost.


carlos8f:
Hi, sorry I was away from Zenbot for so long and never merged this. I've been working on Zenbot 4 and just tagged the release last night. Sim results are now written as a static HTML file. Closing this PR since 3.x won't be supported anymore. Thanks for using Zenbot!

PR #182, opened 2017-05-23
docker-compose: do not expose mongodb by default!

DeviaVir:
This is a pretty important detail for people using docker-compose to deploy their bot on online servers (without default firewall).

There is really no need to expose mongodb at all, since we are internally linking.

I could imagine there are quite a few installations that are currently exposing their mongodb to the outside world, unknowingly.


felipecaldas:
Very much appreciate this pull!!

PR #183, opened 2017-05-23
Docker: extend README slightly

DeviaVir:
Giving a handy reference to people that exclusively use docker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,8 +85,8 @@ docker-compose up (-d if you don't want to see the log)
 If you wish to run commands (e.g. backfills, list-selectors), you can run this separate command after a successful `docker-compose up -d`:
 
 ```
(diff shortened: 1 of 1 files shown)


felipecaldas:
oh wow, champion!

egorbenko:
@DeviaVir add --rm flag to remove container on exit, otherwise you end up with bunch of dangling containers that aren't doing anything

DeviaVir:
@egorbenko good point; I usually run spotify's `docker-gc` every now and then, but this is nice and tidy.

egorbenko:
@DeviaVir i'll let @carlos8f review and merge, looks good to me. thank you for contributing!

carlos8f:
ok cool 👍

PR #195, opened 2017-05-27
docker-compose: Restart node

DeviaVir:
I didn't keep an eye on my bot for a few days only to discover it died a painful death (503 by GDAX being down) without restarting.
Adding `restart: always` should tell docker(-compose) to restart the docker if it stopped for whatever reason.

There might be a case to be made to have `restart: on-failure` but I couldn't think of one.


carlos8f:
thanks

Issue #204, opened 2017-05-31
Add new GDAX selectors to the bot (ETH-EUR and LTC-EUR)


vitorego:
Hi,
There is two new options on GDAX, ETH-EUR and LTC-EUR. Could be nice we you could add this two new selectors,

vitorego:
I just modify the ETH/LTC-USD combinations to EUR inside products.json file and it seems to work.
Is only this file or there is anything else?

micronax:
Try using `extensions/gdax/update-products.sh`

vitorego:
I got this error:
update-products.sh: 2: update-products.sh: Syntax error: "(" unexpected

DeviaVir:
@vitorego 
```
npm install
node extensions/gdax/update-products.sh
```

vitorego:
it works now.
Thanks

PR #205, opened 2017-05-31
GDAX: now supporting EUR/ETH EUR/LTC

DeviaVir:
fixes https://github.com/carlos8f/zenbot/issues/204

### DIFF
--- a/extensions/gdax/products.json
+++ b/extensions/gdax/products.json
@@ -1,65 +1,74 @@
 [
   {
-    "id": "BTC-GBP",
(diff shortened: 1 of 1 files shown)


extensions/gdax/products.json
```diff
@@ -1,65 +1,74 @@
 [
   {
-    "id": "BTC-GBP",
```
DeviaVir, line 3: removing, as the update script does not include them, and I couldn't find any references to this id being in use anywhere.

PR #209, opened 2017-05-31
Implement a basic MACD strategy

dulacp:
This strategy is very simple, once MACD is computed.

My experiments so far, show that it can be very effective for trading periods of 1h, with a shorter period like 15m it seems too erratic and the Moving Averages are kind of lost. But with 1h period it's performing quite well, even in down trending markets.

Also, it's not firing multiple 'buy' or 'sold' signals, only one per trend, which seems to lead to a better quality trading scheme. Especially when the bot will enter in the middle of a trend, it avoids buying unless it's the beginning of the trend.

I've added a little RSI overbought selling signal too, in case `RSI > 70` is true.

Glad to hear your suggestions,
Cheers

### DIFF
--- a/extensions/macd/_codemap.js
+++ b/extensions/macd/_codemap.js
@@ -1,0 +1,6 @@
+module.exports = {
+  _ns: 'zenbot',
+
+  'strategies.macd': require('./strategy'),
+  'strategies.list[]': '#strategies.macd'
+}
--- a/extensions/macd/strategy.js
+++ b/extensions/macd/strategy.js
@@ -1,0 +1,85 @@
+var z = require('zero-fill')
+  , n = require('numbro')
+
+module.exports = function container (get, set, clear) {
+  return {
+    name: 'macd',
+    description: 'Buy when (MACD - Signal > 0) and sell when (MACD - Signal < 0).',
+
+    getOptions: function () {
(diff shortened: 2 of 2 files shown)


DeviaVir (approved):
it might be helpful to update the `README.md` to mention this new strategy as well.

The code definitely looks great and having tested this in live trading it seems to be very clever. I'm keeping this PR open for a bit more visibility from other folks that might know more about MACD (I would expect more indicators, for one). If there is no feedback after a few more days I'll merge your PR.

felipecaldas:
I am adding --strategy macd to the command line but it's not working. Says the strategy does not exist. What am I missing?

DeviaVir:
@felipecaldas don't want to sound mean, but this is not the place for questions like that :) You can create an issue if this really bothers you.
(some simple things you can check; did you check out @dulaccc's branch before attempting to use the strategy?)

felipecaldas:
@DeviaVir Can't find anything there that would allow me to use the strategy. Surely I am missing it, just don't know where to look.
Thanks anyway.

DeviaVir:
Actually, @dulaccc, I think it makes more sense to merge your PR as-is now (I will update README) so more people will play with it and can create issues/PR's easier.

PR #211, opened 2017-06-03
Bittrex: extension support

DeviaVir:
This PR adds support for the bittrex api:
https://bittrex.com/home/api

There's a big limitation in how their API has been adjusted recently, it means we cannot get historical data. I've attempted to make it as clear as possible everywhere that paper/live trading is the only option for now (with bittrex).

screenshot_2017-06-03_09-32-00

### DIFF
--- a/commands/backfill.js
+++ b/commands/backfill.js
@@ -113,7 +113,7 @@ module.exports = function container (get, set, clear) {
                 console.error('\nerror: getTrades() returned duplicate results')
                 console.error(opts)
                 console.error(last_batch_opts)
-                process.exit(1)
+                process.exit(0)
               }
               last_batch_id = trades[0].trade_id
               var tasks = trades.map(function (trade) {
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Zenbot is a hobby project for me and I'm sorry that I can't devote myself full-t
 Zenbot is a command-line cryptocurrency trading bot using Node.js and MongoDB. It features:
 
 - Fully-automated [technical-analysis](http://stockcharts.com/school/doku.php?id=chart_school:technical_indicators:introduction_to_technical_indicators_and_oscillators)-based trading approach
-- Full support for [GDAX](https://gdax.com/), [Poloniex](https://poloniex.com) and [Kraken](https://kraken.com/), work on further exchange support is ongoing.
+- Full support for [GDAX](https://gdax.com/), [Poloniex](https://poloniex.com), [Kraken](https://kraken.com/) and [Bittrex](https://bittrex.com/), work on further exchange support is ongoing.
 - Plugin architecture for implementing exchange support, or writing new strategies
 - Simulator for [Backtesting strategies](https://gist.github.com/carlos8f/b09a734cf626ffb9bb3bcb1ca35f3db4) against historical data
 - "Paper" trading mode, operates on a simulated balance while watching the live market
--- a/conf-sample.js
+++ b/conf-sample.js
@@ -35,6 +35,15 @@ c.kraken.secret = 'YOUR-SECRET'
 // Please read API TOS on https://www.kraken.com/u/settings/api
 c.kraken.tosagree = 'disagree'
 
+// to enable Bittrex trading, enter your API credentials:
+c.bittrex = {}
+c.bittrex.key = 'YOUR-API-KEY'
+c.bittrex.secret = 'YOUR-SECRET'
+// make sure to give your API key access to only: "Trade Limit" and "Read Info",
+// please note that this might change in the future.
+// please note that bittrex API is limited, you cannot use backfills or sims (paper/live trading only)
+
+
 // Optional stop-order triggers:
 
 // sell if price drops below this % of bought price (0 to disable)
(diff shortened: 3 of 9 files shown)


DeviaVir:
@Alexandre-io @nibzo1 @crubb would you mind reviewing this PR (if you have the time), since you are the authors of the Kraken API extension and are (next to Carlos) Knowledgeable People™?

extensions/bittrex/exchange.js
```diff
@@ -0,0 +1,289 @@
...
+ ** Bittrex API
+ ** 
+ ** please note: unfortunately getmarkethistory does not work based on time
+ ** this means that we cannot do any backfilles, but paper trading should be fine
+ ** https://github.com/n0mad01/node.bittrex.api/issues/17
+ **
+ **/
+module.exports = function container(get, set, clear) {
+  var c = get('conf')
+  var recoverableErrors = new RegExp(/(ESOCKETTIMEDOUT|ETIMEDOUT|ECONNRESET|ECONNREFUSED|ENOTFOUND|Invalid nonce|Rate limit exceeded)/)
```
crubb: Isn't `nonce` and `Rate limit` specific to Kraken?
Is there something comparable to `nonce`in Bittrex?

extensions/bittrex/exchange.js
```diff
@@ -0,0 +1,289 @@
...
+        }
+
+        var order = {
+          id: data && data.result ? data.result.uuid : null,
+          status: 'open',
+          price: opts.price,
+          size: opts.size,
+          post_only: !!opts.post_only,
+          created_at: new Date().getTime(),
+          filled_size: '0',
```
crubb: Shouldn't this just be an integer (well, float)? i.e. 0 without quotes?

extensions/bittrex/exchange.js
```diff
@@ -0,0 +1,289 @@
...
+    'apikey' : c.bittrex.key.trim(),
+    'apisecret' : c.bittrex.secret.trim()
+  })
+
+  function joinProduct(product_id) {
+    return product_id.split('-')[0] + '-' + product_id.split('-')[1]
+  }
+
+  function retry(method, args, error) {
+    if (error.message.match(/Rate limit exceeded/)) {
```
crubb: Kraken specific?

extensions/bittrex/exchange.js
```diff
@@ -0,0 +1,289 @@
...
+    trade: function (type, opts, cb) {
+      var args = [].slice.call(arguments)
+
+      var params = {
+        market: joinProduct(opts.product_id),
+        quantity: opts.size,
+        rate: opts.price
+      }
+
+      if(!'ordertype' in opts) {
```
crubb: Is this specific to the bittrex extension? Isn't there some efforts for market vs. limit orders on a higher level?

crubb (commented):
Without actually trying the API here some minor comments.

Why is there a `.jshintrc` in the PR? Could this be messy for other developers?

extensions/bittrex/exchange.js
DeviaVir: Correct, but they do have a `nonce` (which they don't enforce yet, hence the generic message). I don't think they have rate limiting yet, which is why I picked a generic message in case they add it later down the line.

extensions/bittrex/exchange.js
DeviaVir: `API:Rate limit exceeded` is Kraken specific, I picked this generic message in case they decide to rate limit in the future (as they should, really).

extensions/bittrex/exchange.js
DeviaVir: There is, I wanted to be ready for that future by supporting an `ordertype` opt.

edit: what I'm trying to say is, even if the higher level will support it one day, the exchanges will have to support it as well.

extensions/bittrex/exchange.js
DeviaVir: Maybe? I shamelessly copied this over from https://github.com/carlos8f/zenbot/blob/master/extensions/kraken/exchange.js#L200

DeviaVir:
@crubb you're right about the `.jshintrc`, I added it to see if my IDE would be happy, but it was a failed experiment really. Removed it.

extensions/bittrex/exchange.js
crubb: Which we found in https://github.com/carlos8f/zenbot/blob/master/extensions/poloniex/exchange.js#L183 *cough* :D

commands/backfill.js
```diff
@@ -113,7 +113,7 @@ module.exports = function container (get, set, clear) {
                 console.error('\nerror: getTrades() returned duplicate results')
                 console.error(opts)
                 console.error(last_batch_opts)
-                process.exit(1)
+                process.exit(0)
```
DeviaVir, line 116: I'm wondering if this is a good idea:
bittrex doesn't support historical/specific fetching of data, so we need to fetch every minute, but this can still mean we get duplicates (or miss data), exiting with `0` means that we still continue with running paper/live trading.

DeviaVir:
A few days have passed, I think at this point it  makes more sense to merge it and let people play with it (thus getting more feedback on it) then to let it collect dust here.

PR #219, opened 2017-06-04
Tiny Fix: Do not show currencies twice in verbose sell/buy logs

crubb:


lib/engine.js
```diff
@@ -77,7 +77,7 @@ module.exports = function container (get, set, clear) {
     function fc (amt, omit_currency, color_trick, do_pad) {
       var str
       if (isFiat()) {
-        str = n(amt).format('0.00') + (omit_currency ? '' : ' ' + s.currency)
+        str = n(amt).format('0.00')
```
DeviaVir, line 80: for those looking along, this `+ [..]` already happens on https://github.com/crubb/zenbot/blob/4b26487f00312bd9b7ef88f207825723c99d4ab7/lib/engine.js#L96 making this duplicate.

DeviaVir (approved):
great catch!

PR #222, opened 2017-06-05
Mongo: add replicaSet support

### LINKED ISSUE #179: allow for Mongo Replicaset
I use Zenbot dockerized in Kubernetes. Have installed via Helm Mongo-replicaset support. I have it running, but had to change the boot.js file to:
`var u = 'mongodb://' + c.mongo.host + ':' + c.mongo.port + '/' + c.mongo.db + '?replicaSet=rs0'`

If possible it wouldbe great to have an option in the conf.js file to have a replicaset configurable

DeviaVir:
fixes #179

### DIFF
--- a/conf-sample.js
+++ b/conf-sample.js
@@ -8,7 +8,7 @@ c.mongo.db = 'zenbot4'
 c.mongo.username = null
 c.mongo.password = null
 // when using mongodb replication, i.e. when running a mongodb cluster, you can define your replication set here; when you are not using replication (most of the users), just set it to '' (default).
-c.mongo.replicaSet = ''
+c.mongo.replicaSet = null
 
 // default selector. only used if omitting [selector] argument from a command.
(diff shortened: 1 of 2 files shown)


conf-sample.js
```diff
@@ -7,6 +7,7 @@ c.mongo.port = 27017
 c.mongo.db = 'zenbot4'
 c.mongo.username = null
 c.mongo.password = null
+c.mongo.replicaSet = ''
```
crubb: Maybe add a explanatory comment here, as this setting is not self explanatory.

DeviaVir: fair, I added a comment hopefully explaining it and the implementation of it enough.

Fjuxx:
why use `c.mongo.replicaSet = ''` while you are using null for the other fields?
like: `c.mongo.password = null`

PR #223, opened 2017-06-05
Fix up/downtrend thresholds MACD

Fjuxx:
The thresholds were only reacting on the first period of the MACD trend. After this they did not respond as the last period was already positive or negative.

With this it will take the threshold in consideration as well.

### DIFF
--- a/extensions/macd/strategy.js
+++ b/extensions/macd/strategy.js
@@ -52,9 +52,9 @@ module.exports = function container (get, set, clear) {
       }
 
       if (typeof s.period.macd_histogram === 'number' && typeof s.lookback[0].macd_histogram === 'number') {
-        if (s.period.macd_histogram > s.options.up_trend_threshold && s.lookback[0].macd_histogram <= 0) {
+        if (s.period.macd_histogram > s.options.up_trend_threshold && (s.lookback[0].macd_histogram - s.options.up_trend_threshold) <= 0) {
           s.signal = 'buy';
-        } else if (s.period.macd_histogram < -s.options.down_trend_threshold && s.lookback[0].macd_histogram >= 0) {
+        } else if (s.period.macd_histogram < -s.options.down_trend_threshold && (s.lookback[0].macd_histogram + s.options.down_trend_threshold) >= 0) {
           s.signal = 'sell';
         } else {
           s.signal = null;  // hold
--- a/extensions/macd/strategy.js
+++ b/extensions/macd/strategy.js
@@ -52,9 +52,9 @@ module.exports = function container (get, set, clear) {
       }
 
       if (typeof s.period.macd_histogram === 'number' && typeof s.lookback[0].macd_histogram === 'number') {
-        if (s.period.macd_histogram > s.options.up_trend_threshold && (s.lookback[0].macd_histogram - s.options.up_trend_threshold) <= 0) {
+        if ((s.period.macd_histogram - s.options.up_trend_threshold) > 0 && (s.lookback[0].macd_histogram - s.options.up_trend_threshold) <= 0) {
           s.signal = 'buy';
-        } else if (s.period.macd_histogram < -s.options.down_trend_threshold && (s.lookback[0].macd_histogram + s.options.down_trend_threshold) >= 0) {
+        } else if ((s.period.macd_histogram + s.options.down_trend_threshold) < 0 && (s.lookback[0].macd_histogram + s.options.down_trend_threshold) >= 0) {
           s.signal = 'sell';
         } else {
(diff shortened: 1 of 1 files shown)


crubb:
@dulaccc, could you review this PR?

dulacp (changes requested):
Just one suggestion, for the next developers coming across this strategy, can you change:

`if (s.period.macd_histogram > s.options.up_trend_threshold && (s.lookback[0].macd_histogram - s.options.up_trend_threshold) <= 0)` 

to: 

`if ((s.period.macd_histogram -
 s.options.up_trend_threshold) > 0 && (s.lookback[0].macd_histogram - s.options.up_trend_threshold) <= 0)`


And the other condition too, so that we have a clear condition of the form: `if (a - t > 0 && b - t <= 0)`

Fjuxx:
Like that? or did you mean it different?

Ps. wouldn't it be a better option for the strategy always give a buy/sell signal?

Or perhaps the bot telling the strategy that it already bought or sold? as all of the " && (s.lookback[0].macd_histogram - s.options.up_trend_threshold) <= 0)" is only used for not spamming buy/sell signals.
and I've seen other settings/strategies having issues with this.

dulacp:
Exactly, that what I meant 👍
You spotted the correct logic, I think now with this syntax future developers will better understand the conditions. Thanks again for the fix!

PS: I used these conditions to _try_ to make this strategy safer.
Let's imagine that buy signals are always fired if `macd_histo > 0`, and you launch your bot during an up-trend. Then the bot will immediately buy because of the `macd_histo > 0`, but you might be on the end of this up-trend who knows. And if it was the end of it, then you will probably loose money on this trade.

Same thing for the sell signal, you might be on the end of a down trend when running the bot in live trading.

So to _try_ to make it safer, I decided to only trigger buy on the beginning of an up-trend, and sell on the end.

Did you test other conditions? Maybe it performs better?

dulacp:
Thanks for the merge @carlos8f 🍻

Fjuxx:
oh no, i messed it up again i think....
that was my merge into my own master trying to get it up to date.....

Can this PR still be saved or shall i start a new one?

crubb:
Overall still only one file is changed, I think that should be fine.

DeviaVir:
@Fjuxx did you have a rebuttal for @dulaccc's notes:

> So to try to make it safer, I decided to only trigger buy on the beginning of an up-trend, and sell on the end.
> Did you test other conditions? Maybe it performs better?

I'm wondering if this change is "better" or worse. ;)

Fjuxx:
@DeviaVir and @dulaccc 

i did not test with other conditions, as i first wanted to get the initial MACD working. and in my case that meant adding some thresholds to it.

Perhaps it would be best to keep this PR to fixing the thresholds, and perhaps later test some other conditions and add them later if possible.

IMO i think that the thresholds are a must for the MACD as it sometimes is just floating around 0. you don't want to buy/sell every single time it crosses over then. but as soon as it goes up/down more then you want to buy/sell

for now i would recommend to leave it as it is in the PR.

DeviaVir:
@dulaccc what do you think, does this PR make your strategy safer/better, or would it be safer/better to leave as is and perhaps make this adjustable via i.e. command line arguments? (you are still as "requesting changes" on this PR)

Fjuxx:
@DeviaVir 
i could be missing the point here but @dulaccc  and i were discussing about removing the lookback condition. but as he pointed out correctly this would mess if you start the bot in a down or uptrend.

so leaving them in place would bring us back to the original PR

dulacp (approved):

DeviaVir (approved):
:+1:

dulacp:
@DeviaVir this PR fixes only the logic behind the threshold, it's not changing the strategy, neither safer or riskier. So it's perfect to integrate it.

DeviaVir:
@dulaccc got ya! Thanks @Fjuxx and @dulaccc

dulacp:
To reply to your following point @Fjuxx:

> IMO i think that the thresholds are a must for the MACD as it sometimes is just floating around 0. you don't want to buy/sell every single time it crosses over then. but as soon as it goes up/down more then you want to buy/sell

Actually, MACD is not the best for the threshold use cause it's not normalized, so your thresholds will be valid only for a certain period of time, and for a certain volatility.

I'm working on the PPO strategy now, which would be better suited for thresholds usage (I guess)

Cheers

vitoo:
@dulacp Hi Pierre 

Did you succed with PPO threshold ? 
Its better than MACD cross only but still get some false signal depending on trend strengh

PR #226, opened 2017-06-05
Remove full memory dump for Loss/slippage protection

Fjuxx:
Added Regex filter for 'nice errors'.

only filters slippage and loss protection errors to stop a full memory dump. any other error will still do a full memory dump.

Filtered errors will output 1 line with err.message and err.desc:
slippage protection: refusing to buy at 203.12 ZEUR ZEUR, slippage of +0.0%


DeviaVir (approved):
:+1:

PR #228, opened 2017-06-06
Adjusted glob (more structure in extensions possible)

Fjuxx:
Adjusted the glob pattern to search recursive and already search for the _codemap.js file.

This made the following function a lot simpler as glob does al the work for us.

With this change the structure inside the \extensions\ folder does not matter anymore.

please note: I did not include the structure change in the extensions folder yet. as it might mess up the git if they are changed before this goes through.
i would recommend one of the maintainers change the file structure after this is merged. 
This solution does not break the current structure.

### DIFF
--- a/boot.js
+++ b/boot.js
@@ -18,14 +18,12 @@ module.exports = function (cb) {
   zenbot.set('@zenbot:conf', c)
 
   function withMongo () {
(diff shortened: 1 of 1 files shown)


DeviaVir (approved):
smart, :+1:

PR #230, opened 2017-06-06
Extensions: reorder

DeviaVir:
Followup from https://github.com/carlos8f/zenbot/pull/228


Fjuxx:
I've pulled this PR and tested sims, list and trade commands.

they all worked (with different strategies and currencies)

PR #231, opened 2017-06-06
Don't default to live trading

### LINKED ISSUE #232: Change docker-compose to not start in live trading mode by default.
Following the README instructions getting started it is not clear that running docker-compose up will fire up the bot in live trading mode and could be a real surprise to new users starting things up for the first time.

We should have docker configured by default to use the paper mode, and when new users are ready to trade with real currencies adjust it.

Crim:
Following the README instructions getting started it is not clear that running docker-compose up will fire up the bot in live trading mode and could be a real surprise to new users starting things up for the first time.

This PR defaults it to start in paper mode.

Fixes #232

### DIFF
--- a/commands/trade.js
+++ b/commands/trade.js
@@ -106,7 +106,7 @@ module.exports = function container (get, set, clear) {
                 if (so.mode === 'paper') {
                   console.log('!!! Paper mode enabled. No real trades are performed until you remove --paper from the startup command.')
                 }
-		engine.syncBalance(function (err) {
+                engine.syncBalance(function (err) {
                   if (err) {
                     if (err.desc) console.error(err.desc)
                     if (err.body) console.error(err.body)
--- a/commands/trade.js
+++ b/commands/trade.js
@@ -103,7 +103,10 @@ module.exports = function container (get, set, clear) {
               if (err) throw err
               if (!trades.length) {
                 console.log('---------------------------- STARTING ' + so.mode.toUpperCase() + ' TRADING ----------------------------')
-                engine.syncBalance(function (err) {
+                if (so.mode === 'paper') {
+                  console.log('!!! Paper mode enabled. No real trades are performed until you remove --paper from the startup command.')
+                }
+		engine.syncBalance(function (err) {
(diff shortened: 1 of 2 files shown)


DeviaVir:
Valid point, people don't really read before they start. It might make sense, in this case, to add a pretty fat warning (i.e. around here https://github.com/carlos8f/zenbot/blob/master/commands/trade.js#L54) that we are running in paper mode and that the bot is not actually trading unless a user is to remove the `--paper` (after running a bunch of sims, of course...)

DeviaVir (changes requested):
Valid point, people don't really read before they start. It might make sense, in this case, to add a pretty fat warning (i.e. around here https://github.com/carlos8f/zenbot/blob/master/commands/trade.js#L54) that we are running in paper mode and that the bot is not actually trading unless a user is to remove the --paper (after running a bunch of sims, of course...)

Crim:
@DeviaVir Isn't that what this does?  https://github.com/carlos8f/zenbot/blob/master/commands/trade.js#L105 

If you think it still warrants adding an additional log line after that explaining paper mode lemme know and I can add one in 👍

DeviaVir:
@Crim it is, but trying to prevent "my bot is not trading" issues; problem with my suggestion is that there is too much happening there already for a new user to spot those log lines (log lines > importing trades > starting)

It might make sense to add an additional log line at the location you pointed out (https://github.com/carlos8f/zenbot/blob/master/commands/trade.js#L105) if we're in paper mode, something along the lines of: `paper mode enabled, no real trades are made until you remove --paper from the startup command`

or.. thoughts?

Crim:
I bet you're right regarding users submitting issues on that.  I'll add an additional log line :)

Crim:
Updated PR with an additional log when paper is enabled.  Log looks like below:

`server_1        | ---------------------------- STARTING PAPER TRADING ----------------------------`
`server_1        | !!! Paper mode enabled. No real trades are performed until you remove --paper from the startup command.`

DeviaVir (approved):

PR #233, opened 2017-06-06
Order type: allow consumer to specify type of order to use

### LINKED ISSUE #220: Add Market order (instead of limit)
It would be great to have a market order function in the bot.

As it seems that limit orders rarley get filled, and if they get filled it is to late.
and at this moment only a few exchanges have free limit orders. most of them have a fee for all of them, so there is no real good reason to use limit orders on those exchanges.

i looked to update this myself. but i think it will quite a change as zenbot also checks if the limit order is filled or not. not sure how that will work with market orders.

DeviaVir:
Fixes #220

Allow a consumer to pass `--order_type` to `trade` and `sim`, it specifies how to make orders. See the updated `README.md` for more info.

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,7 @@ zenbot trade --help
     -h, --help                      output usage information
     --conf <path>                   path to optional conf overrides file
     --strategy <name>               strategy to use
+    --order_type <type>             order type to use (maker/taker)
     --paper                         use paper trading mode (no real trades will take place)
     --currency_capital <amount>     for paper trading, amount of start capital in currency
     --asset_capital <amount>        for paper trading, amount of start capital in asset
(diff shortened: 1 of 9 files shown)


tuxitor:
Could have updated conf-sample.js with c.order_type default value as well
The addition of order_type is welcome and appreciated.

DeviaVir:
@eigilb good point, updated.

vitorego:
According with gdax website the maximum taker fee is 0.25% and not 0.3%
https://www.gdax.com/fees/BTC-USD
https://www.gdax.com/fees/BTC-EUR

DeviaVir:
@vitorego there is no support for multiple fees at this time, so I took the maximum. (ETH and LTC are at 0.3%)

vitorego:
Ok. By the way, the simulation will include the fees on profit calculation?

DeviaVir:
@vitorego it should, yep. It uses `engine.update` which uses the same functions as I updated here https://github.com/carlos8f/zenbot/pull/233/files#diff-a41b519e7979a713e894ef8d1b3fe568R473

sim results:

`./zenbot.sh sim --order_type=maker --days=7 `

```
end balance: 1178.58656402 (17.86%)
buy hold: 1283.09319115 (28.31%)
vs. buy hold: -8.14%
45 trades over 7 days (avg 6.43 trades/day)
win/loss: 12/33
error rate: 73.33%
```

`./zenbot.sh sim --order_type=taker --days=7 `

```
end balance: 1032.27556848 (3.23%)
buy hold: 1283.09319115 (28.31%)
vs. buy hold: -19.55%
45 trades over 7 days (avg 6.43 trades/day)
win/loss: 12/33
error rate: 73.33%
```

The lower end balance is to be expected, since fees are much higher.

vitorego:
@DeviaVir I got very profitable simulations using short periods and low ema, however in live trading the outcome is not same, even is not profitable at all. I guess with maker type orders the live trading will become profitable on short periods an low ema.

DeviaVir:
@vitorego makes sense, in live trading and limit orders you have to deal with slippage and filling delays. With maker orders the sim will be more close to reality, but it will not be profitable on short periods and low ema (as that results in more trades, more trades results in more fees), I would actually suggest higher periods and higher EMA.

DeviaVir:
Question for those more knowledgeable than me on the subject: if we use market orders, should we be using a different buy and sell price? Assuming we're right now either above or below it to successfully be a market maker.

nedievas:
Maybe it's possible to take dynamic user fees from exchange, like [Bitfinex - api.bitfinex.com/v1/account_infos](https://api.bitfinex.com/v1/account_infos)?

DeviaVir:
@nedievas certainly a good idea, I created https://github.com/carlos8f/zenbot/issues/237

DeviaVir:
Currently running this in a live test, if it does what it's supposed to do I'll merge (unless anyone else has any objects, of course).

vitorego:
I know this pull request was already merged. But I got this idea of having a auto mode on the other type. The boy will start with taker orders (no fee) if the slippage goes over a configured value the bot will close the order using the maker type order. What do you think?

PR #236, opened 2017-06-07
API rate limit issues: c.order_poll_time was not honored in trade.js/engine.js

crubb:
Without the options flag being defined, `order_poll_time` was `undef` in engine.js, therefore calling `checkOrder` too frequently, resulting in API rate limit issues.


DeviaVir (approved):
:+1:

PR #243, opened 2017-06-08
New command "zenbot balance <selector>"

tuxitor:
Introduced new command "zenbot balance <selector>"
Sample output:
BTC-USD Asset: 9.87654321 Currency: 1.23


DeviaVir (approved):
this is already pretty sweet.

I'm wondering if we could have an argument here that would translate current assets to currency, so you can translate to fiat easily. Not sure if that's easy to add (for a later PR/issue) or if you want to add that in this PR.

PR #270, opened 2017-06-12
Bitfinex hotfix

DeviaVir:
fixes #262

PR #273, opened 2017-06-12
Fix #270 and corrected balance calculation

nedievas:
tested and working maker and taker when executing zenbot buy or sell. No more API down!


DeviaVir (approved):
:+1:

PR #274, opened 2017-06-12
Implement version check

### LINKED ISSUE #259: Implement version check
Make sure people are at the very least running node 6 or exit.

DeviaVir:
Fixes #259

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "poloniex.js": "0.0.7",
     "run-parallel": "^1.1.6",
     "run-series": "^1.1.4",
+    "semver": "^5.3.0",
     "sosa_mongo": "^1.0.3",
     "timebucket": "^0.4.0",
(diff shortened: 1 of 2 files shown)


DeviaVir:
@carlos8f is ^ 6.0.0 the version we want to support, or is that a higher/lower scope?

egorbenko:
can we use `process.version` instead of adding another package?

DeviaVir:
@egorbenko this actually uses `process.version`, but a standard version package (semver), which is already installed by other packages as a dependency, is used to do the compare.
https://github.com/carlos8f/zenbot/pull/274/files#diff-aa5e31044d72d2583b98d7a040daca84R9

(hence also no update to `package-lock.json`)

DeviaVir:
this is an issue that keeps popping up in the discord channel, merging this in now to prevent those..

Issue #257, opened 2017-06-10
I got error err.match is not a function in bitfinex/exchange.js


nao0811ta:
> /home/ec2-user/workspace/zen4
> -bit-2/extensions/exchanges/bitfinex/exchange.js:148
>         if (err && err.match(/Error: Invalid order: not enough exchange balance$/)) {
>                        ^
> 
> TypeError: err.match is not a function
>     at /home/ec2-user/workspace/zen4-bit-2/extensions/exchanges/bitfinex/exchange.js:148:24
>     at Request.request [as _callback] (/home/ec2-user/workspace/zen4-bit-2/node_modules/bitfinex-api-node/rest.js:61:14)
>     at Request.self.callback (/home/ec2-user/workspace/zen4-bit-2/node_modules/request/request.js:188:22)
>     at emitTwo (events.js:106:13)
>     at Request.emit (events.js:194:7)
>     at Request.<anonymous> (/home/ec2-user/workspace/zen4-bit-2/node_modules/request/request.js:1171:10)
>     at emitOne (events.js:96:13)
>     at Request.emit (events.js:191:7)
>     at IncomingMessage.<anonymous> (/home/ec2-user/workspace/zen4-bit-2/node_modules/request/request.js:1091:12)
>     at Object.onceWrapper (events.js:293:19)
>     at emitNone (events.js:91:20)
>     at IncomingMessage.emit (events.js:188:7)
>     at endReadableNT (_stream_readable.js:975:12)
>     at _combinedTickCallback (internal/process/next_tick.js:80:11)
>     at process._tickCallback (internal/process/next_tick.js:104:9)

DeviaVir:
What version of `nodejs` are you running? We advise at least 6.x

nao0811ta:
v7.10.0

DeviaVir:
@nedievas any chance you can take a look into this?

nao0811ta:
maybe 
It think that have to fix to 
if (err && err.match && err.match(/Error: Invalid order: not enough exchange balance$/)) {
or 
if (err && String(err).match(/Error: Invalid order: not enough exchange balance$/)) {

nedievas:
Hi, I will. I was at the coast this weekend and see that my bot exited on this error.

nedievas:
WTF. I had working bitfinex extension, and now it's dead... Will revise and update.

nao0811ta:
Mine is working well..

PR #276, opened 2017-06-12
Implement Backfill bitfinex with APIv2 & refactoring err.match function

nao0811ta:
fixed [#257](https://github.com/carlos8f/zenbot/issues/257)


commands/backfill.js
```diff
@@ -66,13 +66,22 @@ module.exports = function container (get, set, clear) {
             var opts = {product_id: product_id}
             if (mode === 'backward') {
               opts.to = marker.from
+              // bitfinex v2 api required timestamp
+              if (exchange_id === 'bitfinex' && !marker.from){
```
DeviaVir, line 70: We will really need to find a different solution for this, logic for exchanges should be limited to exchanges, and never live outside of that.

commands/backfill.js
```diff
@@ -66,13 +66,22 @@ module.exports = function container (get, set, clear) {
...
             }
             else {
               if (marker.to) opts.from = marker.to + 1
               else opts.from = exchange.getCursor(start_time)
             }
             last_batch_opts = opts
-            exchange.getTrades(opts, function (err, trades) {
+            var tradeFunc = 'getTrades'
+            // for bitfinex v2 api for support historical api 
+            if (exchange_id === 'bitfinex'){
```
DeviaVir, line 81: Same comment as before: this logic should happen in the exchange.js

extensions/exchanges/bitfinex/exchange.js
```diff
@@ -22,24 +22,42 @@ module.exports = function container (get, set, clear) {
...
+    return public_client2
+  }
+ 
   function joinProduct (product_id) {
     return product_id.split('-')[0] + '' + product_id.split('-')[1]
   }
 
+  // symbol for v2 
+  function joinProduct2 (product_id) {
+    return 't' + product_id.split('-')[0] + '' + product_id.split('-')[1]
```
DeviaVir, line 38: why not just `return 't' + joinProduct(product_id)`?

extensions/exchanges/bitfinex/exchange.js
```diff
@@ -66,6 +84,48 @@ module.exports = function container (get, set, clear) {
...
+      args.sort = 1
+      // v2 limit is 1000
+      args.limit = 1000
+      if (opts.from) {
+        // initial value is null
+        args.start = opts.from
+      }
+      if (opts.to) {
+        // started point
+        args.end = opts.to
```
DeviaVir, line 101: in the commands you are setting this to `new Date().getTime()`, why not just do that here?

DeviaVir (changes requested):

DeviaVir:
I also think there must be a nicer way than to use `function2` for the v2 things, maybe they should just live in different class?

commands/backfill.js
nao0811ta: Thank you
if opts.to is null , bitfinex response 1 years ago data.
hm.. I'm thinking..

commands/backfill.js
nao0811ta: got it..

extensions/exchanges/bitfinex/exchange.js
nao0811ta: I thought that. I forget to fix.

nao0811ta:
@DeviaVir 
I'm sorry. and thank you, it must create new v2 class. but I don't have many time to survey v2 new API. so I only implemented backfill function for me ASAP. So, I hope that helps someone. 
@nedievas

nao0811ta:
@nedievas  is working for v2
and my implementation is bad 
so close PR.

PR #278, opened 2017-06-12
Unify developers: editorconfig + eslintrc

DeviaVir:
`.editorconfig`:
makes sure every developer's IDE uses the same settings

`.eslintrc`:
makes sure every developer's lint checks against the same settings

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,23 @@
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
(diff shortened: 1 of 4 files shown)


egorbenko:
how about using something like https://houndci.com/ as well?

DeviaVir:
or travis.. or any other CI out there.. @carlos8f dropping a note for you, as only you could enable any of these CI's :)

PR #301, opened 2017-06-16
Add Simulation Timestamp and Directory

codiwest:
I thought it would be nice to have the selector, date, and time appended to the sim_result.html filename. This way the results would not get rewritten for each new simulation. To avoid cluttering the main directory, I added a "simulations" sub-directory.


DeviaVir (approved):
:+1: good idea

PR #313, opened 2017-06-19
Update conf-sample.js : simple warning for sim accuracy

kuhugo:
Added this warning to line 69 & 71. Sim seems to give non-repeatable results when using 100% balance. 

(WARNING : sim won't work properly if you set this value to 100)


DeviaVir (commented):
I think it's a good change, but I wonder if the underlying problem isn't something that would need to be fixed.. @carlos8f might be able to answer that better

codiwest:
Good call @kuhugo !

DeviaVir:
Adding this in for now.

PR #340, opened 2017-06-27
Update exchange.js

kuhugo:
buy / sell orders will cancel themselves after 1 day.


DeviaVir (changes requested):
I would expect this is an optional parameter that you can define yourself, not as a static thing set for everyone.

txssseal:
+1 to config

DeviaVir:
Supserseded by https://github.com/carlos8f/zenbot/pull/596

PR #346, opened 2017-07-01
Fix backtester when run on Windows

cmroche:
Simply needed the 'node' executable added to the command line call.

This calling convention should work on all OSes


DeviaVir:
This change is included with https://github.com/carlos8f/zenbot/pull/344

Issue #352, opened 2017-07-04
Kraken - Unable to trade against EOS currency pair's


nibzo1:
Ive been trying to get Zenbot running against Kraken API with the EOS-ETH currency pair. 

I have ran the Kraken Update products script which has correctly loaded the new currency pairs, although exceptions occur when trying to run the sim or trade. 

Strangle backfill works correctly...

I had the same issue in the past with GNO in the past if i recall? Anyone know if this is a Kraken thing or a Zenbot thing?
```
:~/zbot$ zenbot sim kraken.EOS-ETH
error: could not find product "EOS-ETH"
:~/zbot$ zenbot sim kraken.XEOS-XETH
error: could not find product "XEOS-XETH"
```

nibzo1:
im not sure this fixes the issue, i had previously updated the EOS-ETH currency pair in the `products.json`

backfill worked but trade and sim do not. It may be a kraken issue??

DeviaVir:
@nibzo1 what is the error you are running in to?

nibzo1:
@DeviaVir  been at work all day and only getting to reply now

so I get when i run
```
zenbot backfill kraken.EOS-ETH

skipping 75 hrs of previously collected data

kraken.EOS-ETH saved 1000 trades 1 days left
.....
kraken.EOS-ETH saved 5107 trades 0 days left

download complete!
```

when I run (also same for trade)

```
zenbot sim kraken.EOS-ETH
error: could not find product "EOS-ETH"
```

PR #354, opened 2017-07-05
Update Kraken currency pairs

DeviaVir:
Fixes https://github.com/carlos8f/zenbot/issues/352


nibzo1:
Im pretty sure this doesnt fix anything. I have updated the currency pairs already and it still says it cant finx the currency pair.

Do you know what the pair should look like for the sim command

Issue #357, opened 2017-07-06
Typo in README.md, line 301


tuxitor:
Line 301 should read:
- Pressing `b` will trigger a buy, `s` for sell, and `B` and `S` for market (taker) orders.
```
While the `trade` command is running, Zenbot will respond to these keypress commands:

- Pressing `b` will trigger a buy, `s` for sell, and `B` and `C` for market (taker) orders.
- Pressing `c` or `C` will cancel any active orders.
- Pressing `m` or `M` will toggle manual mode (`--manual`)
```

nik-418:
Added a pull request: https://github.com/carlos8f/zenbot/pull/358

PR #358, opened 2017-07-06
Typo in README.md

nik-418:
I think you meant S instead of C, since C is for cancelling as the line underneath explains.


tuxitor:
Already reported in #357

PR #363, opened 2017-07-07
Fix sma

fasheng:
<del>I hacked the code and generate MACD gnulot data for simulations locally, the result graph just looks same with the online market views :)</del>

Edit: Only fix sma now or it will make MACD strategy works bad..


fasheng:
Hi @talvasconcelos, please take a review work, sir. Hope I didn't misunderstand the code.
https://www.tradingview.com/wiki/Moving_Average#CALCULATION

talvasconcelos:
review pr #369

PR #373, opened 2017-07-09
fix command execution line

nedievas:
```node zenbot``` throws error.
Corrected.

### DIFF
--- a/scripts/auto_backtester/backtester.js
+++ b/scripts/auto_backtester/backtester.js
@@ -79,7 +79,7 @@ let runCommand = (strategy, cb) => {
     speed: `--baseline_periods=${strategy.baseline_periods} --trigger_factor=${strategy.trigger_factor}`,
     trend_ema: `--trend_ema=${strategy.trend_ema} --oversold_rsi=${strategy.oversold_rsi} --oversold_rsi_periods=${strategy.oversold_rsi_periods} --neutral_rate=${strategy.neutral_rate}`
   };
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: scripts/auto_backtester/backtester.js [hunk 1]
```diff
@@ -79,7 +79,7 @@ let runCommand = (strategy, cb) => {
     speed: `--baseline_periods=${strategy.baseline_periods} --trigger_factor=${strategy.trigger_factor}`,
     trend_ema: `--trend_ema=${strategy.trend_ema} --oversold_rsi=${strategy.oversold_rsi} --oversold_rsi_periods=${strategy.oversold_rsi_periods} --neutral_rate=${strategy.neutral_rate}`
   };
-  let command = `node zenbot sim ${simArgs} ${strategyArgs[strategyName]} --period=${strategy.period}  --min_periods=${strategy.min_periods}`;
+  let command = `zenbot sim ${simArgs} ${strategyArgs[strategyName]} --period=${strategy.period}  --min_periods=${strategy.min_periods}`;
```
DeviaVir, line 82: @fasheng added this `node` here. He might have a better opinion on this.

What error do you get, @nedievas?

fasheng: Well, looks the prefix `node` was used to fix Windows platform issue and it was not patched by me(I don't use Windows any more)...
https://github.com/carlos8f/zenbot/pull/346

Anyway, this will break backtester.js under Linux and looks could not solve the Windows issue, too. As stackoverflow discussed, `npm link` should already make a `zenbot.cmd` wrapper for users, so could someone using Windows to test this problem again to ensure if `node zenbot` is necessary.
https://stackoverflow.com/questions/10396305/npm-package-bin-script-for-windows

PR #375, opened 2017-07-10
Strategy: CCI + Stochasctics RSI

talvasconcelos:
Another strategy for testing. Buy when trend is up and sell when it's down. Trend is given by a 50 period EMA. Wanted to get a way for identifying when market is moving sideways by having the change from the EMA be less than a determined threshold but couldn't get it to work for every time period.

Any critics or comments are welcome.


DeviaVir (approved):
:+1: I think it's best to get this out in the wild and allow people to play and extensively test your strats, hence I will merge this now instead of keep this in limbo here.

Thanks for the work and sharing your strategies!

PR #376, opened 2017-07-10
backtester: Rename rsi_macd to srsi_macd

fasheng:


DeviaVir (approved):

talvasconcelos:
Hey @fasheng how do i run the script for srsi_macd strategy? or the script runs all strats?

fasheng:
The following code should works(only test on Linux, but Windows should works, too #373):
```
npm link
./scripts/auto_backtester/backtester.js gdax.BTC-CNY --strategy srsi_macd
# if you run under ./scripts/auto_backtester/, just create empty directory `simulations` manually
```

No, I'm afraid you must edit backtester.js to custom you preferred option ranges step by step to avoid too long testing time:
 https://github.com/carlos8f/zenbot/blob/master/scripts/auto_backtester/backtester.js#L186

PR #380, opened 2017-07-11
Quick fix for the manual process on a remote server

dulacp:
The line [trade.js#L157](https://github.com/carlos8f/zenbot/blob/7829edfde5795781b97a954db70d04bb440c60de/commands/trade.js#L157) was making the script crash on my remote server.

I traced it back to the "remote" console not supporting the raw mode.
But I don't have any knowledge on how to fix it besides the ugly if condition I used.

If you have a suggestion, I'm opened to it.

Cheers


crubb (approved):
"works for me!" ;)

DeviaVir (approved):

PR #386, opened 2017-07-15
Fixes for Quadriga CX trading

cmroche:
* Add amount for takers fee
* Minor update to backfill warning text
* Fix variable shadowing error
* Trades with the same timestamp are now ordered oldest to newest when r eturned from getTrades
* Trade volume now returned as number and not a string, prevents NaN on volume report
* Filter results on timestamp start to prevent re-reporting the same trades in the results


DeviaVir (approved):
:+1:

Issue #385, opened 2017-07-15
Implement mkmarek/forex.analytics: Genetic analysis to build decision trees for buy and sell signals


crubb:
Hey guys,

just documenting @infatrus idea from #general/Discord. https://github.com/mkmarek/forex.analytics looks very promising:

> The result of technical analysis are two binary trees describing strategies for buy and sell signals which produced profit in a certain period of time specified by the input OHLC data set.

@arpheno, @gelotus, what do you guys think of the approach?

Best,
-crubb

Ekliptor:
Great idea.
But can we also use this to optimize parameters of our own strategies written in javascript?
The way I see it this forex depends on the c++ indicators of ta-lib and we can not add our own there? (at least not so easily)

crubb:
That's been done (or is underway at least). Take a look at #298 and #299.

gelotus:
The best person to talk about is @arpheno, he write a thesis about genetics algo.
anyway that lib seems very wip.. And it's for fiat market, will work well in crypto market?
If you are in a good point to merge that with zenbot, simply test it and see the results, for me it's a coin flip

crubb:
zenbot already has some strategies using ta-lib, so that should be fine. Might be working well, see #389 ;)

crubb:
Ok, the current implementation does show some promise (see sim results in the PR #389).

There is still work to do, though (@mkmarek, if you read this, is your forex.analytics still actively maintained?):

- Currently, only the indicators in https://github.com/mkmarek/forex.analytics/tree/master/src/indicators are available (so no EMA, for example).
- Also, the time periods for the indicators are fixed, for example at [14 for RSI](https://github.com/mkmarek/forex.analytics/blob/master/src/indicators/RSI.cpp#L10). This should be optimised by a genetic algorithm as well, I think.

crubb:
Pinged the author, let's see.

sandroid10:
how do i pull this branch ?/

crubb:
Pull zenbot from my github profile and then switch branches to crubb/forex.analytics.

sandroid1:
@crubb when i switch crubb/forex.analytics, i dont see train.js in commands

sandroid1:
@crubb error in this branch 


/home/ubu/mywork/zenbot# ./zenbot.sh train gdax.ETH-USD --days_training 5 --days_test 3 --period 10m --generationCount 2000  --populationCount 200 --indicators CCI,MACD,RSI,SAR
[...]

PR #389, opened 2017-07-16
Add strategy: forex.analytics, an genetic optimization algorithm for TA-lib stats

crubb:
See #385 and see https://github.com/mkmarek/forex.analytics

Furthermore:
- Add option to `sim` to disable output of option
- Add option to `sim` to disable writing HTML results
- Add a 'noop' strategy, which just does nothing. It is used to generate candlestick data for model training.

It seems to work, **but I have no idea what the proper default training options or candlestick length periods should be**.

I have not yet done any in-depth training and optimizations, but out of the box and in the current market conditions at least I get some % above buy/hold ;)

First, you have to backfill, then you train a model, e.g.
`$ ./zenbot.sh train kraken.XETH-ZEUR --days_training 28 --days_test 7 --period 20m`

Running `train` automatically runs simulations on the training and test periods (`--days_test` can be `0` (default)) and outputs the results.

You can then use the model automatically saved in `models/` for trading
`$ ./zenbot.sh trade kraken.XETH-ZEUR --modelfile models/forex.model_kraken.XETH-ZEUR_period-20m_from-20170704_020000_to-20170709_020000.json --period 20m --paper`

Again, this needs optimization on the default parameters and I would love to get feedback from knowledgeable people (@arpheno, @gelotus, @infatrus, ...?).

-- crubb


tuxitor:
Could separating data fetching and training be an idea? Reason is the exchanges that don't provide trade history, like Bitstamp, BTCe and probably more. Then data fetching (possibly without any screen output) could be running uninterupted even if trade stops for shorter or longer periods. Program name could be *fill.js*  as an alternative to *backfill.js*

crubb:
As discussed on #developers/Discord:

This is a general problem for these exchanges, since simulations will also not work properly in the moment, just `--paper` trading. We should be able to play around with this strategy in the other exchanges. If we achieve stellar performance, we should piece together some kind of database-filling method to train the model for these exchanges.

Thank you!

crubb:
Some example ;)

bildschirmfoto 2017-07-17 um 21 51 12

Running `--paper` since a few hours, so far -1.1%/-1.9%, though :/

infatrus:
Model files are overwritten each time, when training with the same parameters. It would be nice to have sim/test results as pct. gain/loose as a part of a filename - that would solve the issue with overwriting, and would let easily find the best model.

crubb:
D'oh.. That does explain some of the variance in my results :D
Fixed.

infatrus:
;) Yeah, same with me - I had some promising model for BTC-USDT, then just realized, that files are overwritten... Retard me... I could make a copy.

extensions/strategies/forex_analytics/strategy.js
```diff
@@ -0,0 +1,87 @@
+var z = require('zero-fill')
```
txssseal, line 1: would be really awesome to get a brief commented rundown of what this strategy does.

infatrus:
crubb, can you add "--reverse" option to allow trades with reversed buy/sell signals?
Some models could probably deal better with such option.

2017-07-21 05:00:00  0.00000717 STR-BTC   -0.8%    2.18m   57 ++      0.1106   0.0337      buy
2017-07-21 05:10:00  0.00000701 STR-BTC   -2.2%    4.91m     - 46    -0.0653   0.0415     sell
2017-07-21 05:20:00  0.00000722 STR-BTC   +3.0%    4.45m   57 ++      0.1596   0.0752      buy

sandroid10:
t

sandroid10:
@crubb . i am getting following error. 

/home/ubu/newbot/zenbot# ./zenbot.sh train gdax.XETH-ZEUR --days_training 28 --days_test 7 --period 20m
/home/ubu/newbot/zenbot/node_modules/codemap/index.js:372
          throw err;
          ^

Error: cannot alter undefined path `zenbot:conf`
    at /home/ubu/newbot/zenbot/node_modules/codemap/index.js:369:21
    at Array.forEach (native)
    at Object.validatePathCache (/home/ubu/newbot/zenbot/node_modules/codemap/index.js:302:35)
    at Object.set (/home/ubu/newbot/zenbot/node_modules/codemap/index.js:194:13)
    at PlainObject.set (/home/ubu/newbot/zenbot/node_modules/codemap/index.js:66:22)
    at module.exports (/home/ubu/newbot/zenbot/boot.js:18:10)
    at Object.<anonymous> (/home/ubu/newbot/zenbot/zenbot.sh:20:1)
    at Module._compile (module.js:569:30)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:503:32)

commands/sim.js
```diff
@@ -63,6 +64,7 @@ module.exports = function container (get, set, clear) {
           so.start = d.subtract(so.days).toMilliseconds()
         }
         so.stats = !!cmd.enable_stats
+        so.show_options = !!!cmd.disable_options
```
DeviaVir: TIL :)

DeviaVir (approved):
The only real comment I have is that I would expect an entry in the `/docs/` on how to use the training model, at least with a simple usage example. Other than that I think this code is fine and is ready to be used by more people so we can get more eyes and comments on it.

crubb:
Documentation added.

PR #392, opened 2017-07-19
Quadriga CX now supports LTC trading, adding to products

cmroche:


cmroche:
Also added a few more bug fixes related to passing numbers as strings.

extensions/exchanges/quadriga/products.json
```diff
@@ -34,5 +33,14 @@
...
     "label": "ETH/BTC"
+  },
+  {
+    "id": "LTCCAD",
+    "asset": "LTC",
+    "currency": "CAD",
+    "min_size": "0.00001",
+    "max_size": "10000",
+    "increment": "0.00001",
+    "label": "LTC/CAD"
```
DeviaVir, line 44: is it really just LTC/CAD? No other pairs?

DeviaVir (approved):

extensions/exchanges/quadriga/products.json
cmroche: Yeah. QCX only does CAD currency and no other crypto-to-crypto trading other than ETH-BTC for the moment. But they are fiarly new, so that may change with time.

Issue #390, opened 2017-07-17
Bittrex:  TypeError: Cannot read property 'match' of undefined


webworker01:
Running MACD paper test on Bittrex
```
/zenbot/extensions/exchanges/bittrex/exchange.js:77
          if (data.message.match(recoverableErrors)) {
                          ^

TypeError: Cannot read property 'match' of undefined
    at /root/zenbot/extensions/exchanges/bittrex/exchange.js:77:27
    at Request._callback (/root/zenbot/node_modules/node.bittrex.api/node.bittrex.api.js:112:25)
    at self.callback (/root/zenbot/node_modules/request/request.js:188:22)
    at emitOne (events.js:115:13)
    at Request.emit (events.js:210:7)
    at Request.onRequestError (/root/zenbot/node_modules/request/request.js:884:8)
    at emitOne (events.js:115:13)
    at ClientRequest.emit (events.js:210:7)
    at TLSSocket.socketErrorListener (_http_client.js:399:9)
    at emitOne (events.js:115:13)
    at TLSSocket.emit (events.js:210:7)
    at emitErrorNT (internal/streams/destroy.js:62:8)
    at _combinedTickCallback (internal/process/next_tick.js:102:11)
    at process._tickCallback (internal/process/next_tick.js:161:9)
```

arenstar:
+1 i have this issue also

benpdavison:
Hello, written in some code to hotfix this, i.e. stop the match if message not present. see this PR: https://github.com/carlos8f/zenbot/pull/412

PR #412, opened 2017-07-27
added check for message before doing anything with it

benpdavison:
Fix for this bug: https://github.com/carlos8f/zenbot/issues/390 

Think the main question here is why the MACD paper test on Bittrex is not returning a message. However the check to see if meesage exists should stop it erroring out.


DeviaVir (approved):
fair enough, thanks for the hotfix :+1:

PR #418, opened 2017-07-28
updated c.default_selector to c.selector in README.md

natefanaro:
I think c.default_selector was from v3 and that was changed to c.selector in v4. Updated README.md to reflect that.


DeviaVir (approved):
:+1: good catch

PR #424, opened 2017-07-29
Fix docker builds for forex.analytics

cmroche:
The forex.analytics documentation says to install this package globally, I am unable to test if it work locally since forex.analytics itself does not build with `docker-compose build`


cmroche:
I was able to fully fix the docker build, npm now needs the --unsafe-perm option passed to it as forex.analytic needs this to properly call it's post-install scripts.

DeviaVir (approved):

PR #426, opened 2017-07-30
Support for missing win/loss and error rate from output

natefanaro:
The output does not include win/loss and error rate when no trades have been performed. Seen here https://github.com/carlos8f/zenbot/blob/master/commands/sim.js#L137

This commit checks to see if the regex match is empty first before attempting to use it as an array.


DeviaVir (approved):
:+1: thanks for taking the time to submit this

PR #438, opened 2017-08-03
Update exchange products for BCH support

cmroche:
Largely this is adding support for BCH
Updated: bitfinex, bittrex, kraken

Poloniex and gdax was updated, but had no changes.
BTC-e could not be updated (broken update script)
Quadriga doesn't support BCH

BCH currency pairs on Kraken don't fully match what is returned by their API. ie: API says BCH-ZUSD, but only accepts BCH-USD. This is manually updated in the product list, I suspect it is a bug that will be rectified on their side in a few days and may require an additional future update.

Tested all pairs via backfill.


DeviaVir (approved):
Thanks!

PR #445, opened 2017-08-06
Add support for Gemini exchange platform

cmroche:
This change principally add support for the Gemini exchange, including the option to switch between the live and sandbox environment. The API is tested via backfill, buy, sell, trade (live and paper in sandbox). Populated products and added an update script.

In the process of implementing the exchange I've identified and corrected a few other bugs and inconsistencies in other platforms along the way. Please note the updates to the developer documentation around subtleties in the way the getCursor function is used by the backfill command. If not implemented properly the `--days` option may be ignored during the backfill.

Also identified and corrected a bug in the way kraken reported rejected orders. It's possible that someone reported rejected orders as done to prevent re-posting of cancelled post-only market orders. Right now there is no option to enable or disable this feature in the engine. If this is the case, then the option should be added to allow users to choose between standard limit, or post-only orders.


DeviaVir (approved):
:+1: great, let's see this in the wild

mkaskov:
Error: Cannot find module 'gemini-api'

vaisov:
> Error: Cannot find module 'gemini-api'

Fixed by running `npm install gemini-api`

Issue #466, opened 2017-08-12
Live trading appears to be broken after this commit.


davidandrewcope:
Live trading appears to be broken after commit af971ef3. I get to ---STARTING LIVE TRADING--- then nothing... Strangely, I can still toggle Manual mode, but nothing else. Rolling back to the previous commit locally takes care of the issue.

gasperzupancic:
i can confirm this situation. Thanks @davidandrewcope for pointing this out. it was making me crazy when i was trying to figure it out yesterday and over night:S

cmroche:
It won't show anything until it receives valid trade data, if you have short periods on a low volume exchange there is no data until the first live trade is reported by the exchange.

Previous behaviour is that it would fail pre_roll, and then fallback to loading an entire day worth of data during the live update stage potentially signalling bad trade orders on the exchange while it effectively caught up.

@davidandrewcope @gasperzupancic 

As an example, kraken.XETH-ZCAD has period of an hour with no volume. I've found that it can take 10-20 minutes often to receive the first sample, then you will begin to see data including 0 volume periods reporting.

Can you please confirm that this is the case you are seeing.

While it may be possible to inject 0 volume periods on the exchange once live trading is started, it isn't a trivial change and will have to wait until I have some time to tackle the problem.

cmroche:
After thinking about the problem a bit, I could probably add a check to detect that we don't gave enough data to start the reports and throw out a warning to the console in stead. I'll look at this in the short term.

cmroche:
https://github.com/carlos8f/zenbot/pull/470

dymex0:
same here, 1h later and my live trading on gdax is still empty

cmroche:
@dymex0 I'll investigate, GDAX uses a reverse history scan, it could be related. But for now can you please tell me the following.

1) Can you reproduce the problem with paper trading?
2) Which currency pairs specifically are you using on GDAX.
3) What parameters are you using for trading (basically strategy, and configuration options for it)

Thanks

cmroche:
I've found the problem, it effects any exchange platform that doesn't use timestamps for filtering/pagination. I'll have a fix in the next 24 hours.

cmroche:
Fixed in pull request #470

PR #470, opened 2017-08-13
Fix live trading for GDAX (and others) add a message to improve low volume pre-roll

cmroche:
This is an improvement to the previous low volume fix. In low volume markets there may not be any trades during the pre_roll or initial live trading periods, it is not possible to show reports in that case which gives the impression that the bot it broken. Now we will at least show a message indicating that we are waiting for data before reporting starts.

Also provide a fix for broken trading GDAX. The exchange doesn't use timestamps so it is necessary to call getCursor when setting the trade_cursor.

### DIFF
--- a/commands/trade.js
+++ b/commands/trade.js
@@ -114,11 +114,10 @@ module.exports = function container (get, set, clear) {
               limit: 1000
             }
             if (db_cursor) {
-              trade_cursor = db_cursor
               opts.query.time = {$gt: db_cursor}
             }
             else {
-              trade_cursor = query_start
+              trade_cursor = s.exchange.getCursor(query_start) 
               opts.query.time = {$gte: query_start}
             }
             get('db.trades').select(opts, function (err, trades) {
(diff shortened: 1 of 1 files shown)


cmroche:
Should address this https://github.com/carlos8f/zenbot/issues/466

commands/trade.js
```diff
@@ -114,11 +114,10 @@ module.exports = function container (get, set, clear) {
               limit: 1000
             }
             if (db_cursor) {
-              trade_cursor = db_cursor
```
cmroche, line 117: This line isn't actually necessary, was added in a previous PR I sent. This conditional branch once getNext is called following any valid data. The function passed to engine.update which set the db_cursor will also properly set the trade_cursor using a call to s.exchange.getCursor as one would expect.

commands/trade.js
```diff
@@ -114,11 +114,10 @@ module.exports = function container (get, set, clear) {
...
               opts.query.time = {$gt: db_cursor}
             }
             else {
-              trade_cursor = query_start
```
cmroche, line 121: This line however is necessary, since we may quit pre-roll after 1 iteration and no data may be present. In that case we would have an invalid trade_cursor once we started updating the engine causing exchange.getTrades to return and process as many results as possible in a single API call.

DeviaVir (approved):

Issue #427, opened 2017-08-01
Kraken updateproducts.sh does not work


smartbrain:
Hi,

The Kraken products list is dated with many pairs that no longer exist (all the ones with the old names that do not begin with x or z).

The updateproducts.sh script does not work properly. Here's the error message I get.

Running Ubuntu 16.04 LTS.

I am otherwise able to trade on Kraken (and bittrex's updateproducts works. Incidentally the permissions on bittrex's updateproducts.sh is incorrect and lacks execute (chmod +x) - or should I open a separate ticket for that)

Thanks and keep up the good work.

```
> don@naked:~/zenbot/extensions/exchanges/kraken$ ./update-products.sh 
Error: Could not understand response from server: <!DOCTYPE html>
<!--[if lt IE 7]> <html class="no-js ie6 oldie" lang="en-US"> <![endif]-->
<!--[if IE 7]>    <html class="no-js ie7 oldie" lang="en-US"> <![endif]-->
<!--[if IE 8]>    <html class="no-js ie8 oldie" lang="en-US"> <![endif]-->
<!--[if gt IE 8]><!--> <html class="no-js" lang="en-US"> <!--<![endif]-->
<head>
<meta http-equiv="set-cookie" content="cf_use_ob=0; expires=Tue, 01-Aug-17 20:38:56 GMT; path=/">
<title>api.kraken.com | 520: Web server is returning an unknown error</title>
<meta charset="UTF-8"/>
<meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
<meta http-equiv="X-UA-Compatible" content="IE=Edge,chrome=1"/>
<meta name="robots" content="noindex, nofollow"/>
<meta name="viewport" content="width=device-width,initial-scale=1,maximum-scale=1"/>
<link rel="stylesheet" id="cf_styles-css" href="/cdn-cgi/styles/cf.errors.css" type="text/css" media="screen,projection"/>
<!--[if lt IE 9]><link rel="stylesheet" id='cf_styles-ie-css' href="/cdn-cgi/styles/cf.errors.ie.css" type="text/css" media="screen,projection" /><![endif]-->
<style type="text/css">body{margin:0;padding:0}</style>
<!--[if lte IE 9]><script type="text/javascript" src="/cdn-cgi/scripts/jquery.min.js"></script><![endif]-->
[...]

PR #483, opened 2017-08-17
FIxing kraken product update and product list

cmroche:
Adds a check for the known broken currencies

### DIFF
--- a/extensions/exchanges/kraken/exchange.js
+++ b/extensions/exchanges/kraken/exchange.js
@@ -38,6 +38,19 @@ module.exports = function container(get, set, clear) {
     return product_id.split('-')[0] + product_id.split('-')[1]
   }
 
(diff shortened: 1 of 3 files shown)


cmroche:
Fixes bug https://github.com/carlos8f/zenbot/issues/427

timstoop:
Tested this locally, but it only fixes part of the problem. With this patch applied and running `balance kraken.BCH-EUR`, it does not recognise the EUR I have in my account.

cmroche:
Ok, thanks for the feedback. I should be able to devise a solution.

cmroche:
@timstoop Try applying the change again, should work now.

timstoop:
balance works, backfill works and trade --paper works! Great stuff!

DeviaVir (approved):

PR #518, opened 2017-09-03
Interactive CUI

nedievas:
Added more simple interaction options and some colors. Some early statistics on exit. Works in paper and live trading. Simple and more elegant?

capture


codiwest:
I think it looks pretty nice.

DeviaVir (approved):

PR #532, opened 2017-09-10
The robo kitten patch

ghost:
Basically self explanatory.

Just a small conf-sample.js note about max sell loss percent.

### DIFF
--- a/conf-sample.js
+++ b/conf-sample.js
@@ -102,7 +102,7 @@ c.buy_pct = 99
 c.sell_pct = 99
 // ms to adjust non-filled order after
 c.order_adjust_time = 5000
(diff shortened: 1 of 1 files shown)


DeviaVir (approved):

crubb:
Hey,

I'm not sure this is correct. This setting seems to specify that no sell is undertaken if the loss on your order is >25% (at least per default config).

It's used here: https://github.com/carlos8f/zenbot/blob/master/lib/engine.js#L458

Setting this to 0 sells even at a massive loss, instead of holding. It has nothing to do with "ensure selling at a higher price" and the comment is misleading.

I would recommend reverting this commit.

Best,
crubb

PR #536, opened 2017-09-10
Make "post_only" orders and option in the config file

cmroche:


cmroche:
I messed this up, will redo.

Issue #439, opened 2017-08-04
: No such file or directoryenv: node on docker


jooshxp:
Running on windows with Docker Toolbox after : docker-compose build and docker-compose --file=docker-compose-windows.yml up this logs shown : 

> Starting zenbot_mongodb-data_1 ...
Starting zenbot_mongodb-data_1 ... done
Starting zenbot_mongodb_1 ...
Starting zenbot_mongodb_1 ... done
Starting zenbot_server_1 ...
Starting zenbot_server_1 ... done
Attaching to zenbot_mongodb-data_1, zenbot_mongodb_1, zenbot_server_1
mongodb_1       | 2017-08-04T05:11:53.423+0000 I CONTROL  [initandlisten] MongoD
B starting : pid=1 port=27017 dbpath=/data/db 64-bit host=6f04527b4111
mongodb_1       | 2017-08-04T05:11:53.423+0000 I CONTROL  [initandlisten] db ver
sion v3.4.6
mongodb_1       | 2017-08-04T05:11:53.423+0000 I CONTROL  [initandlisten] git ve
rsion: c55eb86ef46ee7aede3b1e2a5d184a7df4bfb5b5
mongodb_1       | 2017-08-04T05:11:53.423+0000 I CONTROL  [initandlisten] OpenSS
L version: OpenSSL 1.0.1t  3 May 2016
mongodb_1       | 2017-08-04T05:11:53.423+0000 I CONTROL  [initandlisten] alloca
tor: tcmalloc
mongodb_1       | 2017-08-04T05:11:53.423+0000 I CONTROL  [initandlisten] module
s: none
mongodb_1       | 2017-08-04T05:11:53.423+0000 I CONTROL  [initandlisten] build
environment:
mongodb_1       | 2017-08-04T05:11:53.423+0000 I CONTROL  [initandlisten]     di
stmod: debian81
mongodb_1       | 2017-08-04T05:11:53.423+0000 I CONTROL  [initandlisten]     di
starch: x86_64
mongodb_1       | 2017-08-04T05:11:53.423+0000 I CONTROL  [initandlisten]     ta
rget_arch: x86_64
mongodb_1       | 2017-08-04T05:11:53.423+0000 I CONTROL  [initandlisten] option
s: { storage: { mmapv1: { smallFiles: true } } }
zenbot_mongodb-data_1 exited with code 0
mongodb_1       | 2017-08-04T05:11:53.426+0000 I -        [initandlisten] Detect
[...]

PR #549, opened 2017-09-20
Refactor bootstrap to address issues with windows docker builds

cmroche:
Simple refactor allows bootstrap to direct call the zenbot.js script passing in command line parameters to that script instead of wrapping it in a bash/sh script. This prevent line ending types from preventing the scripts from running when built on windows.

The new system defined a zenbot.bat and zenbot.sh script which wrap their calls and pass all parameters directly to zenbot.sh via node. To the user there should be no difference in how the script is invocated when done locally.


cmroche:
Should resolve issue #439

Dockerfile
```diff
@@ -5,7 +5,11 @@ WORKDIR /app
 
 COPY package.json /app/
 RUN npm install -g node-gyp && npm install --unsafe-perm
+RUN npm install
```
DeviaVir: why again after already running `npm install --unsafe-perm`?

cmroche: Missed that. Should be fixed now.

DeviaVir (approved):

PR #560, opened 2017-09-24
Correcting asset and currency formatting

nedievas:
Before:
```2017-09-24 19:08:31  201.720 ZEC-USD   +0.03%       10   58 ++      0.0219   0.0096           0.00 ZEC  0.14594 USD   -0.17%   +0.41%```
After:
```2017-09-24 19:03:41  200.760 ZEC-USD   -0.02%        6    -- 43    -0.0108   0.0068           0.00000727 ZEC  0.15 USD   -0.17%   +0.89%```

### DIFF
--- a/lib/engine.js
+++ b/lib/engine.js
@@ -79,7 +79,7 @@ module.exports = function container (get, set, clear) {
     }
 
     function isFiat () {
-      return !s.currency.match(/^BTC|ETH|XMR|USD|EUR|USDT$/)
+      return !s.currency.match(/^BTC|ETH|XMR|USDT$/)
     }
 
     var max_fc_width = 0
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by nedievas** [comment]
🤒

File: lib/engine.js [hunk 1]
```diff
@@ -79,7 +79,7 @@ module.exports = function container (get, set, clear) {
     }
 
     function isFiat () {
-      return !s.currency.match(/^BTC|ETH|XMR|USD|EUR|USDT$/)
+      return !s.currency.match(/^BTC|ETH|XMR|USDT$/)
```
DeviaVir, line 82: `USD|EUR` is not fiat? What is the thought process behind this?

nedievas: If we leave it, we get like that: ``` 0.00000 BTC  1000.00000 EUR```
If we return to @carlos8f [origin](https://github.com/carlos8f/zenbot/blob/6162ffdfb72f6542a871a34ae5e76cf8cfb90737/lib/engine.js#L64), we have correctly displayed currencies: ```0.00000 BTC  1000.00 EUR```

DeviaVir (approved):

PR #568, opened 2017-09-26
updates to CCI-SRSI strategy

talvasconcelos:
I messed up the last PR... Now it should update the strategy

Added an acceleration parameter to try to find sideways markets.


DeviaVir (approved):
:+1:

PR #569, opened 2017-09-26
Update README.md

brucetus:
Running commands with docker is different now (since pull #549)
I couldn't use
`docker run --rm --link zenbot_mongodb_1:mongodb -it zenbot_server ./zenbot.sh list-selectors`
which is what the main zenbot instructions say to use.
I had to remove `./zenbot.sh` and it works.
`docker run --rm --link zenbot_mongodb_1:mongodb -it zenbot_server list-selectors`
Maybe we need to change the readme.md to update this.


DeviaVir (approved):
:+1:

PR #572, opened 2017-09-27
missing dep "eyes": "^0.1.8"

grigio:


DeviaVir (changes requested):
You will also need to include an updated `package-lock.json`:

```
rm -rf package-lock.json
npm i
git add package-lock.json
git commit
```

grigio:
Ok I'll update it. I noticed I had to install manually another dependency for the bitfinex support

grigio:
I'm not sure it is needed I tried to clone a new zenbot at it seems to work without it

cmroche:
This is also resolved in PR #576 The problem isn't that eyes is missing from the package.json, but that the package-lock.json wasn't updated after the package "cliff" was added to the project. Eyes is included indirectly from this package, updating the package lock resolves issues with both "cliff" and "eyes".

Issue #455, opened 2017-08-08
Backfill poloniex.* no responding


phongtt3:
Hello,
I can't backfill any pair from poloniex from zenbot. 

When i try to open poloniex's public api from browser, i meet google's image captcha, so i think zenbot can't crawl data because this captcha. 

Is there any work-around way to backfill data from Poloniex? if yes, can you pls tell me.

---

edit by @DeviaVir:

@pushkarnagle + @ivanmladek + @dmzpkts experience the same problem.

For more people experiencing this problem, please use the emoji reacts.

pushkarnagle:
Same error. Can download any pair.

ivanmladek:
Same error.

dmzpkts:
Also have this error.

krull:
@phongtt3 Confirmed that this bug is replicatable:

```
docker run --rm --link zenbot_mongodb_1:mongodb -it zenbot_server ./zenbot.sh backfill poloniex.LTC-BTC --days 90

skipping 184 hrs of previously collected data

poloniex.LTC-BTC saved 200 trades 83 days left
.
error: getTrades() returned duplicate results
{ product_id: 'LTC-BTC', to: 1501709244 }
{ product_id: 'LTC-BTC', to: 1501709244 }
```

Any thing can be done to make this succeed?

ivanmladek:
Replicable.

sudo docker run --rm --link zenbot_mongodb_1:mongodb -it zenbot_server ./zenbot.sh backfill poloniex.GAME-BTC --days 100

poloniex.GAME-BTC saved 200 trades 100 days left
............
poloniex.GAME-BTC saved 14636 trades 99 days left
...........
poloniex.GAME-BTC saved 3036 trades 98 days left
...........
poloniex.GAME-BTC saved 6478 trades 97 days left
...........
poloniex.GAME-BTC saved 3938 trades 96 days left
...........
poloniex.GAME-BTC saved 2536 trades 95 days left
...........
poloniex.GAME-BTC saved 2596 trades 94 days left
...........
poloniex.GAME-BTC saved 3127 trades 93 days left
...........
error: getTrades() returned duplicate results
{ product_id: 'GAME-BTC', to: 1501709452 }
{ product_id: 'GAME-BTC', to: 1501709452 }

diymusic:
Same problem. Replicable.

./zenbot.sh backfill poloniex.dash-btc --days 36
poloniex.DASH-BTC saved 200 trades 6 days left
.
[...]

PR #580, opened 2017-10-03
#455 increase getTrades time window from 2 to 12 hours

ratajczak:


466:
Did not solve the issue of duplicated results for me.
I had to increased the time significantly, up to 200 hours (ok, might have worked somewhere below 200, but i did not want to try dozens of limits....)

PR #333, opened 2017-06-25
xmpp for trading alarms

piedrogenius:
Added optional account set-up and buy/sell alarms with XMPP. Needs simple-xmpp and 2 XMPP server accounts: one for sending the trader messages and one for receiving them.

Considering two way functionality to disable/reenable live trading while running

PR #583, opened 2017-10-03
Added send/buy orders notifications using Slack's Incoming WebHooks API.

flocca:


evseevnn-zz:
@flocca i think bad idea implement it inside engine.
i think that should be like module. Im sure what many ppl wanna write notification module for this project and we need prepare possibility for that.

Please try refactor your code and make it better.

flocca:
Ok. I can code this like a zenbot extension, maybe under extensions/notifications/, a place where we can put all the notifications modules (slack.js, email.js, etc.). What do you think about all this @evseevnn?

evseevnn-zz:
@flocca Yes! That good idea! You will make a great contribution to the development of this library if you do this.

tuxitor:
@flocca 
I agree with @evseevnn in this matter. XMPP has already been added to "lib/engine.js". To continue adding messaging options this way only leads to more code pollution. I recommend reading this thread:
https://github.com/carlos8f/zenbot/pull/333,
and make note of the comments by @DeviaVir and @tuxitor. Then take a look at the "zentalk" branch on this fork where a more generiic approach is used for messaging: https://github.com/tuxitor/zenbot/tree/zentalk
The documentation should be informative: https://github.com/tuxitor/zenbot/blob/zentalk/docs/Zentalk.md
The "zenxmpp" program should make a good base for implementing any messaging client. Here is an example:
zenxmpp

DeviaVir:
I think I'll close this PR for now, the changes are good, but since you have the intention of rewriting it doesn't make too much sense to keep it around.

flocca:
Thanks @DeviaVir, I agree.

PR #596, opened 2017-10-06
GDAX: cancel_after

DeviaVir:
Superseding https://github.com/carlos8f/zenbot/pull/340

Please review:
- [x] @crubb 
- [x] @nedievas

### DIFF
--- a/conf-sample.js
+++ b/conf-sample.js
@@ -96,7 +96,7 @@ c.profit_stop_pct = 1
 
 // avoid trading at a slippage above this pct
 c.max_slippage_pct = 5
-// buy with this % of currency balance (WARNING : sim won't work properly if you set this value to 100) 
+// buy with this % of currency balance (WARNING : sim won't work properly if you set this value to 100)
 c.buy_pct = 99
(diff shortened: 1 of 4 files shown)


nedievas:
Hi. It would be nice to use similar option for Bitfinex by adding manual cancel to replicate this GDAX order parameter, wouldn't?

crubb (approved):

nedievas:
Reading [GDAX API docs](https://docs.gdax.com/#place-a-new-order) I found that we miss another option for 'cancel_after' to work as intended: 
> Time in force
> Time in force policies provide guarantees about the lifetime of an order. There are four policies: good till canceled GTC, good till time GTT, immediate or cancel IOC, and fill or kill FOK.
> GTT Good till time orders remain open on the book until canceled or the allotted cancel_after is depleted on the matching engine. GTT orders are guaranteed to cancel before any other order is processed after the cancel_after timestamp which is returned by the API. A day is considered 24 hours.

Please confirm that this enhancement is really working before merge.

DeviaVir:
@nedievas good point, added.

DeviaVir:
Seems to work. Merging.

PR #299, opened 2017-06-16
Genetic algorithm to fine tune strategy parameters

arphen:
I don't think it's in a state to be merged, but i'd like to have a discussion

PR #598, opened 2017-10-06
Genetic algorithm to fine tune strategy parameters

DeviaVir:
Since https://github.com/carlos8f/zenbot/pull/299 has stalled for quite a while now, I grabbed the code and put it in the `scripts/` dir, so it'll be easier for other devs to keep working on it. In its current state, the code has unfortunately decayed quite a bit.

Please review:
- [x] @crubb

### DIFF
--- a/scripts/genetic_algo/parsing.py
+++ b/scripts/genetic_algo/parsing.py
@@ -1,0 +1,27 @@
+import shlex
+import subprocess
+import re
+
+
+def parse_trades(stuff):
+    """
+    >>> parse_trades("1 trades over 17 days (avg 0.06 trades/day)")
+    '0.06'
+    :param stuff:
+    :return:
+    """
+    return stuff.split(b'avg')[-1].strip().split()[0]
+
+
+def args_for_strategy(strat):
+    ansi_escape = re.compile(b'\x1b[^m]*m')
+    available = subprocess.check_output(shlex.split('/app/zenbot.sh list-strategies'))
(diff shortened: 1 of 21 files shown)


crubb (changes requested):
Good idea! I'd recommend extending README.md with links to the original Github issue, PR and this one with the warning, that the code decayed quite a bit.

scripts/genetic_algo/parsing.py
```diff
@@ -0,0 +1,25 @@
...
+    '0.06'
+    :param stuff:
+    :return:
+    """
+    return stuff.split(b'avg')[-1].strip().split()[0]
+
+
+def args_for_strategy(strat):
+    available = subprocess.check_output(shlex.split('/app/zenbot.sh list-strategies'))
+    strats = [strat.strip() for strat in available.split(b'\n\n')]
```
emabo: `import re`

`ansi_escape = re.compile(b'\x1b[^m]*m')`
`available = subprocess.check_output(shlex.split('/app/zenbot.sh list-strategies'))`
`strats = [ansi_escape.sub(b'', strat.strip()) for strat in available.split(b'\n\n')]`

Could I propose these lines to get rid of ANSI escape sequences? otherwise it seems to me that this function is not able to parse correctly command options. NB: I'm using python scripts outside docker container.

DeviaVir: Added.

PR #603, opened 2017-10-09
Fix calls for running zenbot in genetic algo directory

emabo:
I would like to propose a fix for running genetic algo using python code in scripts/genetic_algo. I'm not sure this fix is going to work also using docker, because I'm running the app without dockers.

### DIFF
--- a/scripts/genetic_algo/evaluation.py
+++ b/scripts/genetic_algo/evaluation.py
@@ -31,7 +31,7 @@ def runzen(cmdline):
 
 
 class Andividual(Individual):
-    BASE_COMMAND = '/app/zenbot.sh sim {instrument} --strategy {strategy} --avg_slippage_pct 0.33 --filename temp.html'
+    BASE_COMMAND = 'env node ../../zenbot.js sim {instrument} --strategy {strategy} --avg_slippage_pct 0.33 --filename temp.html'
     def __init__(self, *args,**kwargs):
         super(Andividual, self).__init__(*args, **kwargs)
         self.args = args_for_strategy(self.strategy)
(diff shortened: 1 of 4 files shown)


scripts/genetic_algo/evaluation.py
```diff
@@ -31,7 +31,7 @@ def runzen(cmdline):
 
 
 class Andividual(Individual):
-    BASE_COMMAND = '/app/zenbot.sh sim {instrument} --strategy {strategy} --avg_slippage_pct 0.33 --filename temp.html'
+    BASE_COMMAND = 'env node ../../zenbot.js sim {instrument} --strategy {strategy} --avg_slippage_pct 0.33 --filename temp.html'
```
DeviaVir, line 34: I understand the `../../` but why can't you use `zenbot.sh`?

emabo: because, otherwise I have this error:


module.js:473
      throw err;
      ^

Error: Cannot find module '/home/emanuele/Code/zenbot.gh/scripts/genetic_algo/zenbot.js'
    at Function.Module._resolveFilename (module.js:527:15)
    at Function.Module._load (module.js:453:25)
    at Function.Module.runMain (module.js:665:10)
    at startup (bootstrap_node.js:201:16)
    at bootstrap_node.js:626:3


it expects to find zenbot.js in current directory that is scripts/genetic_algo/

JensvdHeydt:
I think we should decide if we expect the scripts to be run from the /scripts/ folder of from the root folder. That's the main cause of most of the problems here

DeviaVir (approved):

DeviaVir:
I think the docker variant is also not working now, so better to have a working example then nothing at all. Fully expecting a followup PR for this.

Issue #514, opened 2017-09-01
Kraken AddOrder Issue


ghost:
Unhandeld AddOrder error:
Error: Kraken API returned error: Order:Invalid price:XLTCZUSD price can only be specified up to 4 decimals.
    at Request._callback (/home/masternode/zenbot/node_modules/kraken-api/kraken.js:155:34)
    at Request.self.callback (/home/masternode/zenbot/node_modules/request/request.js:188:22)
    at emitTwo (events.js:125:13)
    at Request.emit (events.js:213:7)
    at Request.<anonymous> (/home/masternode/zenbot/node_modules/request/request.js:1171:10)
    at emitOne (events.js:115:13)
    at Request.emit (events.js:210:7)
    at IncomingMessage.<anonymous> (/home/masternode/zenbot/node_modules/request/request.js:1091:12)
    at Object.onceWrapper (events.js:314:30)
    at emitNone (events.js:110:20)
    at IncomingMessage.emit (events.js:207:7)
    at endReadableNT (_stream_readable.js:1059:12)
    at _combinedTickCallback (internal/process/next_tick.js:138:11)
    at process._tickCallback (internal/process/next_tick.js:180:9)

{ Error: 
order rejected
    at /home/masternode/zenbot/lib/engine.js:233:23
    at KrakenClient.<anonymous> (/home/masternode/zenbot/extensions/exchanges/kraken/exchange.js:275:20)
    at Request._callback (/home/masternode/zenbot/node_modules/kraken-api/kraken.js:155:23)
    at Request.self.callback (/home/masternode/zenbot/node_modules/request/request.js:188:22)
    at emitTwo (events.js:125:13)
    at Request.emit (events.js:213:7)
    at Request.<anonymous> (/home/masternode/zenbot/node_modules/request/request.js:1171:10)
    at emitOne (events.js:115:13)
    at Request.emit (events.js:210:7)
    at IncomingMessage.<anonymous> (/home/masternode/zenbot/node_modules/request/request.js:1091:12)
    at Object.onceWrapper (events.js:314:30)
    at emitNone (events.js:110:20)
    at IncomingMessage.emit (events.js:207:7)
    at endReadableNT (_stream_readable.js:1059:12)
    at _combinedTickCallback (internal/process/next_tick.js:138:11)
    at process._tickCallback (internal/process/next_tick.js:180:9)
  order: 
   { id: null,
     status: 'rejected',
     price: '76.30815150',
     size: '8.80839814',
[...]

PR #616, opened 2017-10-12
Fix kraken products for krakens update on decimal precision

Marbel:
Fix the products for kraken exchange.
There was an update on the kraken exchange for the decimal precision of currency pairs.
I changed the min_size and increment accordingly to the following blog post of kraken [Kraken Blog post on decimal precision](https://blog.kraken.com/post/1278/announcement-reducing-price-precision-round-2/) 

This should fix the following issue #514

### DIFF
--- a/extensions/exchanges/kraken/products.json
+++ b/extensions/exchanges/kraken/products.json
@@ -2,239 +2,239 @@
   {
     "asset": "BCH",
     "currency": "ZEUR",
-    "min_size": "0.01",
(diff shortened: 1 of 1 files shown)


DeviaVir (approved):
Great work! :+1:

MarcoWagner:
Accordingly to the blog post the decimal precision of the price got reduced, not the precision of the asset size. This seems to limit both which makes it impossible to trade small amounts.

PR #632, opened 2017-10-13
Update engine.js

### LINKED ISSUE #628: GDAX - Good Till Time - expire_time required
#596 appears to not include an expire_time variable in the request.

Error

sniper7kills:
Should fix #628


sniper7kills:
disregard

DeviaVir:
@sniper7kills this is in a different place than the other PR, I don't think it hurts to have it there as well

Issue #630, opened 2017-10-13
status report feature during trading 


ituhin:
is there anyway to check the status report while trading is running in real/paper mode.
like the way you can see trade options by pressing "O" 
can we see the status report  without exiting ?

nedievas:
what do you mean by status report? Please, define.

ituhin:
like.... whats the current profit or loss... how many trades it did.... the exact result which it shows when you exit the trading by pressing "X"
I cant find a way to see the statistics without exiting.

nedievas:
These stats you see are currently alpha version as simply moved from sim to live without deeper inspection, and as observed does not represent real trading data. My purpose is to refine it more.

ituhin:
ok. just noticed its not correct.  as it shows its losing against buy-hold with a 0 trade.


Exiting... 
Writing statistics...

last balance: 973.69976804 (-2.63%)
buy hold: 995.44910878 (-0.46%)
vs. buy hold: -2.18%
0 trades over 5 days (avg 0.00 trades/day)

and it was not actually 5 days, it was 20 minutes

emabo:
sorry I did some mess with commits, please look only at latest one.
My intention is to add a new 'P' key to print stats without exiting from zenbot.

mikeh21:
Printing stats while running in --paper mode causes stats to break

[code]
manual limit BUY command executed
2017-10-25 15:14:00  4236.26 BTC-GBP   +0.11%        0           4236.26004233.5800     sell  0.00000 BTC  1000.00 GBP   +0.00%   -0.36
%
2017-10-25 15:14:01  4236.26 BTC-GBP   +0.00%        0           4236.26004233.5800           0.00000 BTC  1000.00 GBP   +0.00%   -0.36
2017-10-25 15:14:01  4236.26 BTC-GBP   +0.00%        0           4236.26004233.5800           0.00000 BTC  1000.00 GBP   +0.00%   -0.36
2017-10-25 15:14:01  4236.26 BTC-GBP   +0.00%        0           4236.26004233.5800           0.00000 BTC  1000.00 GBP   +0.00%   -0.36
2017-10-25 15:14:01  4236.26 BTC-GBP   +0.00%        0           4236.26004233.5800           0.00000 BTC  1000.00 GBP   +0.00%   -0.36
[...]

PR #637, opened 2017-10-14
Print stats while running trade command without exiting (#630)

emabo:
Add new 'P' key for printing stats without stopping the bot.
This could be a fix for issue #630. 
I don't fully understand why at line 136 a new dummy sell op is added. Anyway, this could be fine for printing stats before exiting, but not for printing partial results during running process.
At line 167, probably we can leave only 'sell > 0' condition to avoid division by zero; it should cover all the cases.


DeviaVir (approved):
:+1:

PR #653, opened 2017-10-23
Pulls min_size from Bittrex instead of being a static value.  Creates…

jerrytunin:
… a static value for min_total of 50k Satoshi (0.005BTC), Bittrex's minimum order size.


DeviaVir (approved):
:+1:

PR #656, opened 2017-10-25
Genetic backtester fix

FunFR:
- Remove duplicate args
- Reduce disk usage
-  Move output files from root dir to sim dir


DeviaVir (approved):
:+1:

PR #663, opened 2017-10-28
WIP: docker-compose based command interface

dkarlovi:
Instead of using `docker` with linking and naming specific containers, we just execute the command through `docker-compose` itself.


DeviaVir (approved):
Looks good: is it still WIP?

dkarlovi:
@DeviaVir yes, I'd like to figure out how we can deal with permissions mismatch in `models` and `simulations` when running via Docker (with or without Compose). We can merge this, I'll add an additional PR for that, though.

PR #672, opened 2017-11-01
CEXIO Backfill duplicated entry error fix

nedievas:
should fix backfill bug.
Please, test and confirm. My system passed test.
Essential part of code:
```
getTrades: function (opts, cb) {
      var func_args = [].slice.call(arguments)
      if (typeof opts.from === 'undefined') {
        var args = 1000
      } else {
        args = opts.from
      }
```


DeviaVir (approved):
:+1:

Issue #679, opened 2017-11-03
Stochastic CCI Strategy has no strategy name


agarzon:
Execute ```zenbot list-strategies```  and you get the list of supported strategies, all of them has a canonical one word name, except Stochastic CCI Strategy

```
Stochastic CCI Strategy
  description:
    Stochastic CCI Strategy
  options:
    --period=<value>  period length (default: 20m)
    --min_periods=<value>  min. number of history periods (default: 30)
    --ema_acc=<value>  sideways threshold (0.2-0.4) (default: 0.03)
    --cci_periods=<value>  number of RSI periods (default: 14)
    --rsi_periods=<value>  number of RSI periods (default: 14)
    --srsi_periods=<value>  number of RSI periods (default: 9)
    --srsi_k=<value>  %K line (default: 5)
    --srsi_d=<value>  %D line (default: 3)
    --oversold_rsi=<value>  buy when RSI reaches or drops below this value (default: 18)
    --overbought_rsi=<value>  sell when RSI reaches or goes above this value (default: 85)
    --oversold_cci=<value>  buy when CCI reaches or drops below this value (default: -90)
    --overbought_cci=<value>  sell when CCI reaches or goes above this value (default: 140)
    --constant=<value>  constant (default: 0.015)
If you have questions about this strategy, contact me... @talvasconcelos

forex_analytics
  description:
    Apply the trained forex analytics model.
  options:
    --modelfile=<value>  modelfile (generated by running `train`), should be in models/ (default: none)
    --period=<value>  period length of a candlestick (default: 30m) (default: 30m)
    --min_periods=<value>  min. number of history periods (default: 100)

macd
  description:
    Buy when (MACD - Signal > 0) and sell when (MACD - Signal < 0).
  options:
    --period=<value>  period length (default: 1h)
    --min_periods=<value>  min. number of history periods (default: 52)
    --ema_short_period=<value>  number of periods for the shorter EMA (default: 12)
    --ema_long_period=<value>  number of periods for the longer EMA (default: 26)
    --signal_period=<value>  number of periods for the signal EMA (default: 9)
    --up_trend_threshold=<value>  threshold to trigger a buy signal (default: 0)
    --down_trend_threshold=<value>  threshold to trigger a sold signal (default: 0)
    --overbought_rsi_periods=<value>  number of periods for overbought RSI (default: 25)
    --overbought_rsi=<value>  sold when RSI exceeds this value (default: 70)

rsi
  description:
[...]

PR #680, opened 2017-11-03
Fix issue #679

nedievas:


agarzon:
Jesus Christ... that was fast!

DeviaVir (approved):

PR #707, opened 2017-11-14
Update to HitBTC 2.0 (not fully functional)

brucetus:
Fixed problem with previous hitbtc exchange.js, then updated it to ccxt hitbtc2. Works better than the previous version but buys/sells don't go through. Still working on that. Thanks to @abduegal for finding problems.


extensions/exchanges/hitbtc/exchange.js
```diff
@@ -31,15 +31,15 @@ module.exports = function container (get, set, clear) {
...
+    }, 5000)
   }
 
   var orders = {}
 
   var exchange = {
     name: 'hitbtc',
     historyScan: 'forward',
-    makerFee: 0.01,
+    makerFee: -0.01,
```
DeviaVir, line 42: Really?

DeviaVir (approved):

extensions/exchanges/hitbtc/exchange.js
brucetus: Yessir https://hitbtc.com/fees-and-limits. 0.01% rebate on maker orders

Issue #708, opened 2017-11-14
Bittrex and Insufficient Funds


KryptoNova:
I'm trying to place a manual trade in a live trading environment and am getting the error below.  I have the buy_pct and sell_pct values set to 100 so that the entire balance is bought and sold each time.  I am thinking it has something to do with rounding and it is one Sat more than what is in the account.  I'm not exactly sure where this calculation takes place, but I guess it is in engine.js somewhere.  Could someone point me to where this calculation takes place so I can handle rounding errors so the 100% purchase of asset works?  Thanks!

```
manual limit BUY command executed
INSUFFICIENT_FUNDS


{ Error: order not found in cache
    at Object.getOrder (/root/nodeJS/zenbot/extensions/exchanges/bittrex/exchange.js:263:29)
    at Timeout.checkOrder [as _onTimeout] (/root/nodeJS/zenbot/lib/engine.js:283:24)
    at ontimeout (timers.js:469:11)
    at tryOnTimeout (timers.js:304:5)
    at Timer.listOnTimeout (timers.js:264:5) desc: 'could not execute buy: error placing order' }

```

ether-btc:
with bittrex "99.5"% works. With Gdax I found, "99.6"% works.
`buy_pct=99.5 and buy_pct=99.6` respectively.
`sell_pct=100` works fine both on bittrex and gdax.

Is it the bot that fails to being able to buy 100% or is it the exchanges fault I wonder. Never got to find out who is the culprit.

KryptoNova:
To answer @ether-btc it is both.  The bot wasn't taking a fee into account when determining the amount of asset to buy to leave enough BTC in the account to cover the fee.  I think that Bittrex needs a way to handle a 1 sat rounding problem.

Closing for now as I do believe that #711 has addressed this for 100% buys.

PR #711, opened 2017-11-15
Bittrex and Insufficient Funds (#708)

KryptoNova:
Added if statements so the exchange can have it's cut out of the purchase of the asset.
Most exchanges want their cut in BTC rather than the alt-coin being processed.
This modification allows for the exchange fee to be removed from the overall purchase price allowing for 100% of the currency balance to be spent without an insufficient funds error popping up.

### DIFF
--- a/lib/engine.js
+++ b/lib/engine.js
@@ -413,7 +413,15 @@ module.exports = function container (get, set, clear) {
           if (signal === 'buy') {
             price = n(quote.bid).subtract(n(quote.bid).multiply(so.markup_pct / 100)).format(s.product.increment, Math.floor)
             if (!size) {
-              size = n(s.balance.currency).multiply(so.buy_pct).divide(100).divide(price).format('0.00000000')
+              if (so.mode === 'live') {
+                if (so.order_type === 'maker')
+                  size = n(s.balance.currency).multiply(so.buy_pct).divide(100).multiply(s.exchange.makerFee / 100).format('0.00000000');
+                else
+                  size = n(s.balance.currency).multiply(so.buy_pct).divide(100).multiply(s.exchange.takerFee / 100).format('0.00000000');
+                size = n(s.balance.currency - size).divide(price).format('0.00000000')
+              } else {
+                size = n(s.balance.currency).multiply(so.buy_pct).divide(100).divide(price).format('0.00000000')
+              }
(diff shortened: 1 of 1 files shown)


lib/engine.js
```diff
@@ -413,7 +413,15 @@ module.exports = function container (get, set, clear) {
           if (signal === 'buy') {
             price = n(quote.bid).subtract(n(quote.bid).multiply(so.markup_pct / 100)).format(s.product.increment, Math.floor)
             if (!size) {
-              size = n(s.balance.currency).multiply(so.buy_pct).divide(100).divide(price).format('0.00000000')
+              if (so.mode === 'live') {
+                if (so.order_type === 'maker')
+                  size = n(s.balance.currency).multiply(so.buy_pct).divide(100).multiply(s.exchange.makerFee / 100).format('0.00000000');
+                else
+                  size = n(s.balance.currency).multiply(so.buy_pct).divide(100).multiply(s.exchange.takerFee / 100).format('0.00000000');
+                size = n(s.balance.currency - size).divide(price).format('0.00000000')
```
DeviaVir, line 421: this overwrites the result of your if at :417

KryptoNova: That is correct.  I use "size" as a transitional holding variable in a live run of Zenbot.  This is because there could be two different fees if the bot is running in maker or taker mode with the exchange.

The size variable that is calculated at :418 and :420 represents the amount of the exchange fee that is going to be taken if the purchase is completed after the buy_pct is calculated.

The final size variable at :421 represents the fee adjusted size of the amount of asset to purchase at the selected price. 

This allows enough BTC to remain in the account to allow 100% buy of an asset without having a problem with insufficient funds.  The side effect is that if you use a 50% buy_pct, it will not be the full 50% due to the fee coming out of the size.  This is the formula, where fee represents maker or taker fee:  

`size = (balance * buy_pct - (balance * buy_pct * fee)) / price`

Let me know if this needs more clarification or if there is a clearer way.  This worked really well with the Bittrex exchange and eliminated the Insufficient Funds problems that I was running into.

DeviaVir: Ah, I looked over ` - size` there. Gotcha.

DeviaVir (approved):

PR #715, opened 2017-11-15
Different Bittrex error (#659)

### LINKED ISSUE #659: Different Bittrex error
I already used the fix in #457 and it worked for a couple days. However, I got a new error that occured after selling. 

```
/home/b/zenbot/extensions/exchanges/bittrex/exchange.js:85
        Object.keys(data.result).forEach(function (i) {
               ^

TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at /home/b/zenbot/extensions/exchanges/bittrex/exchange.js:85:16
    at Request._callback (/home/b/zenbot/node_modules/node.bittrex.api/node.bittrex.api.js:114:25)
    at Request.self.callback (/home/b/zenbot/node_modules/request/request.js:188:22)
    at emitTwo (events.js:125:13)
    at Request.emit (events.js:213:7)
    at Request.<anonymous> (/home/b/zenbot/node_modules/request/request.js:1171:10)
    at emitOne (events.js:115:13)
    at Request.emit (events.js:210:7)
    at IncomingMessage.<anonymous> (/home/b/zenbot/node_modules/request/request.js:1091:12)
    at Object.onceWrapper (events.js:314:30)
    at emitNone (events.js:110:20)
    at IncomingMessage.emit (events.js:207:7)
    at endReadableNT (_stream_readable.js:1057:12)
    at _combinedTickCallback (internal/process/next_tick.js:138:11)
    at process._tickCallback (internal/process/next_tick.js:180:9)
```

The sell order already printed too and it said 'sold' for about 10 minutes before I got the error.

KryptoNova:
This fix should handle any error that happens and then fire's off a retry.  I tried the Try .. Catch block on fiddle and it worked flawlessly to capture the error in question.

I'm not sure how to exactly test the retry, but I created the statement similar to the one on line 78.  This technically should be recoverable, because Bittrex is just having some sort of hiccup on their API sending data.

I would say this would close #659, but time will ultimately tell.

### DIFF
--- a/extensions/exchanges/bittrex/exchange.js
+++ b/extensions/exchanges/bittrex/exchange.js
@@ -82,7 +82,7 @@ module.exports = function container(get, set, clear) {
         }
 
         var trades = []
-        if (typeof data.result !== 'undefined') {
+        try {
           Object.keys(data.result).forEach(function (i) {
             var trade = data.result[i]
             trades.push({
@@ -93,6 +93,9 @@ module.exports = function container(get, set, clear) {
               side: trade.OrderType == 'BUY' ? 'buy' : 'sell'
             })
           })
+        } catch (e) {
+          console.log('bittrex API (getmarkethistory). Retry in progress.  Error:' + e);
+          return retry('getTrades', func_args, e.toString());
         }
         cb(null, trades)
       })
--- a/extensions/exchanges/bittrex/exchange.js
+++ b/extensions/exchanges/bittrex/exchange.js
@@ -95,7 +95,7 @@ module.exports = function container(get, set, clear) {
           })
         } catch (e) {
           console.log('bittrex API (getmarkethistory). Retry in progress.  Error:' + e);
-          return retry('getTrades', func_args, e.toString());
+          return retry('getTrades', func_args, {message: e.toString()});
         }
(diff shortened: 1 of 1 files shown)


DeviaVir (approved):
:+1:

ljankok:
Unfortunately the issue is still there.

bittrex

KryptoNova:
@ljankok , could it be that you have the version that I first committed?  I found that using the typeof in the if did not work properly.  I had to convert it to a Try...Catch statement to get this one and work around it.

I was just running it today and it happened.  It seems to have worked for me.
image

If you are comfortable reviewing the code.  Take a look at exchange.js in the bittrex folder.  On line 85 you should see a try statement instead of an if statement.  If you see an if, then you have the first attempt at trying to get this to work properly.

Please let us know if it is a try instead of an if.  It is hard to duplicate this due to the nature of the occurance.  So it has to be dealt with in the wild and I will try to come up with Plan C at this point.

ljankok:
After I have started zenbot again I can also confirm that the fix is working.

I reviewed the code and your catch statement indeed looks like the type of exception which is needed.

bittrex-fix

Issue #724, opened 2017-11-16
buy_pct no longer works


brucetus:
I'm pretty sure this is from commit b53abcc. I posted a comment here https://github.com/carlos8f/zenbot/commit/b53abcc294fa6394d8d3f37c532b7690bc1655e8 that explains more

KryptoNova:
I am still getting the hang of how to use GitHub.  I made the corrections in the wrong place, but I think I have it right now.

I have made the adjustment to the final size calculation so it takes the buy_pct into account prior to the removal of the applicable fee.  So the resulting purchase of the asset will be the most allowed for the buy_pct after the exchange fee is accounted for.  
Doing it this way will eliminate problems with insufficient funds when doing 100% purchases on the exchange.

PR #725, opened 2017-11-17
Fix for buy_pct (#724)

KryptoNova:
This is a fix to address #724.  This was related to commit b53abcc not taking the buy_pct into account during the final calculation of the size.  It was taking it into account when calculating the impact of the fee on the asset purchase, but not the overall purchase of the asset itself.


DeviaVir (approved):
:+1:

Issue #717, opened 2017-11-16
Incorrect Volume Column


KryptoNova:
I've noticed that there is a problem where the Volume column is not correct.  It may only be for exchanges that to not support a time range, like Bittrex.  The problem is that the exchange will report the 200 most recent changes, resulting in duplicates.  Every time another batch of 200 is pulled based on the refresh rate, the amounts applicable to the current period keep adding to the trade volume.  For strategies that use volume as basis for an indicator, this could lead to false signals if the "period" is set high.

I'm working on a pull request that fixes this.  However, I have run into a little problem when Zenbot starts.  It appears that during the pre-roll, it will retrieve the 200 most recent trades.  After the history is completed, Zenbot will then query for the 200 most recent trades again.  This causes the current row on the screen output to show a doubled volume.

It appears that the index that I am using to try to eliminate duplicate trade id numbers is being reset to an empty array after the pre-roll completes.  Is there a way to have a persistent variable that has scope within the exchange.js file?

KryptoNova:
I think I figured out the problem.  I was able to fix the Bittrex exchange adapter so that it does not double the volume levels anymore.  I'm going to let Zenbot run for a while and prove out some of the results to real market activity.  If all goes well, I will create a pull request for it.  

BTW, this fix also appears to have fixed the duplicate record error that the code warns about and I will remove that warning as part of the pull request.

PR #729, opened 2017-11-18
Fix Volume in Bittrex Adapter

KryptoNova:
This corrects the volume reporting in the Bittrex adapter.  It also removes the need to have the warning about the duplicate results as that is no longer a problem.

Adjust the wording in the error message so the console.log was removed.  Utilize the retry function's error reporting to show the message instead.

If this is approved, then #717 can be closed.


DeviaVir (approved):
LGTM

PR #730, opened 2017-11-18
Fix poll_trades From Command Line

KryptoNova:
When using the poll_trades setting from the command line, Zenbot will ignore it in favor of the currently used conf.js file.  This effectively eliminates the possibility of an override from the command line as dictated in the settings.

This modification will allow the command line option to override the setting in the conf.js file.


DeviaVir (approved):
Good catch :+1:

PR #735, opened 2017-11-19
Kraken network errors

xma:
This pull request intend to fix two things, for kraken exchange:
- Errors returned by Kraken (w/ Cloudflare proxy) weren't handled, namely:
   SSL handshake failed
   Gateway time-out
   Bad gateway

It's quite bad as those could occur while doing an order or a cancel
which in this case wouldn't have the 'retry' feature

- Timeout, related to the NPM module "kraken-api"
It's set to 5 sec, which is way too low ( I'm seeing sometime requests completed in around 30s, especially those related to order)

=> I've made a quick fork of the module, setting the default to 60s

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "glob": "^7.1.1",
     "har-validator": "^5.0.3",
     "idgen": "^2.0.2",
-    "kraken-api": "^0.1.7",
+    "kraken-api": "github:xma/npm-kraken-api#zenbot",
     "micro-request": "^666.0.10",
     "mime": "^1.4.0",
     "minimist": "^1.2.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -31,7 +31,7 @@
     "glob": "^7.1.1",
     "har-validator": "^5.0.3",
     "idgen": "^2.0.2",
-    "kraken-api": "^0.1.7",
+    "kraken-api": "github:xma/npm-kraken-api#zenbot",
```
DeviaVir, line 34: Can you explain why we need to jump to your fork for this over waiting for your change to be pulled in there?

xma:
I'll test if I can propagate a timeout value, from conf.js to kraken-api
If that works it'll remove dependency to my fork

anyway commit 0d3b901d7c3f304d1230fb033c8d64c2cdd3253f could be integrated

MT00x:
Great ! Should kraken-api be update to 1.0.0 ?

xma:
Maybe
in kraken-api 1.0.0 there's now a handy constructor for setting timeout

but now it's using await/async & promise, so it's more work to convert current code here

PR #737, opened 2017-11-21
Added currency to the notifications via bots

mxaddict:
Added currency to the notifications via bots IE: discord/xmpp


DeviaVir (approved):
:+1:

PR #740, opened 2017-11-23
Updated the bitfinex product settings for XMR/BTC

mxaddict:
Updated the bitfinex product settings for XMR/BTC

### DIFF
--- a/extensions/exchanges/bitfinex/products.json
+++ b/extensions/exchanges/bitfinex/products.json
@@ -98,9 +98,9 @@
   {
     "asset": "XMR",
     "currency": "BTC",
-    "min_size": "0.1",
+    "min_size": "0.2",
     "max_size": "5000.0",
-    "increment": "0.0001",
+    "increment": "0.000001",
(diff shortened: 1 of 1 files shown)


nedievas:
I think there is no need to update separate pairs every time Bitfinex changes minimal order. You can pull these by executing `update-producs.sh`. Increment is a pain in the ***, and these should be updated to everyone's personal favor. I get better ROI when trading LTC-BTC with increment 0.0001, worse when increment is 0.0000001.

mxaddict:
@nedievas 
Hmmm, you think that a larger increment is better?

Do you mind sharing your base conf for doing LTC-BTC, I want to tune my XMR-BTC trading and make some OK profit, then I can invest more into my bot's balance :)

DeviaVir:
I agree with @nedievas here, if anyone would PR a latest `update-products.sh` result I'll accept it.

mxaddict:
Created PR #744 as suggested/requested :)

PR #766, opened 2017-11-29
More accurate simulation

FunFR:
I think the changes below make the simulation more accurate
- "order.price" instead or "orig_price" to compute sim price
- slippage only for taker order

Try and tell me


brucetus:
Slippage should only be for maker orders

FunFR:
I prefer compute the last limit order price which match market price instead of applying slippage to orig_price

DeviaVir (approved):
I agree with your thinking here.

PR #774, opened 2017-11-30
Update package.json

ghost:
For reverting to v2 api


ghost:
I'm working on v3 api but it will require code edits. This is a temporary workaround so their exchange won't crash? Note. this version actually works.

DeviaVir (commented):
I'm not sure how this is a downgrade?

ghost:
In our npm module, I had v3 api, but we decided to switch to v2 api. I was making v2 code api calls to their v3 api in our npm module. (which doesnt work lol) So I bumped version of the npm module that way the crash won't happen again.

ghost:
You can see what changed in the module here: https://github.com/Gr1dd/node-wexnz/commit/5ac123646afa6a9e5e741e6866cd6dec6dc092f4#diff-d59ec7e466cd48ceb2729ca30a2c300e

PR #785, opened 2017-12-03
Add two commands to output live HTML chart without killing zenbot. Either on demand or automatically.

tiagosiebler:
• Add command to write html chart of trades so far, without needing to kill bot via X command. (lower-case d)
• Add command to enable saving/dumping live charts to file every X seconds (defaulted at 10 seconds), relative to selector & year + month + date. (upper-case D to toggle on/off)


tfossoulw:
Didn't review but great functionnality :+1:

DeviaVir (approved):
:+1: nice!

PR #791, opened 2017-12-03
You'll never find out what files are in this folder !

storm1er:
Just add the folder `/conf` in `.gitignore` so we can keep a clean file tree =)


ArnaudRitti:
Integrated in PR #850 ,  i use the same thing

storm1er:
@ArnaudRitti Thanks ^^

Duplicate of #850

PR #797, opened 2017-12-04
Rest API Support for zenbot

abduegal:
Hi guys,

Here is the code, that enables zenbot to have a REST API.
I tried to keep the code extendable, by allowing zenbot to add multiple **outputs** (one of them is this REST API). I have other **outputs** available like storing the trade in AWS dynamoDB. 
I also think it would be a good idea to refactor the stdout reporting part of zenbot, so that would also become an **output**.

Please let me know if you have any questions.


storm1er:
Can you give us an exemple about how to use your output extension ? =)
_The same question goes for the REST API. (README.md appending ?)_

This PR feels like "+16843214 up" ... You could split this in some PR next time ^^'

extensions/output/api.js
```diff
@@ -0,0 +1,34 @@
...
+module.exports = function container (get) {
+  let c = get('conf')
+  let express = require('express')
+  let app = express()
+  let random_port = require('random-port')
+
+  let run = function(reporter, tradeObject) {
+    if (!reporter.port || reporter.port === 0) {
+      random_port({from: 20000}, function(port) {
+        startServer(port, tradeObject)
```
DeviaVir, line 12: try/catch if the port doesn't work, and we try a different one automatically?

extensions/output/api.js
```diff
@@ -0,0 +1,34 @@
...
+      })
+    } else {
+      startServer(reporter.port, tradeObject)
+    }
+  }
+
+  let startServer = function(port, tradeObject) {
+    tradeObject.port = port
+
+    app.get('/tradedata', function (req, res) {
```
DeviaVir: `/trades`?

extensions/output/api.js
```diff
@@ -0,0 +1,34 @@
...
+      res.send(tradeObject)
+    })
+
+    app.listen(port)
+    tradeObject.url = require('ip').address() + ':' + port + '/tradeData'
```
DeviaVir: `/trades`?

DeviaVir (commented):
I'm not sure if `lib/output.js` is the right place for this, but have no other suggestions at this time. I'll let others take a look and am interested in their feedback

extensions/output/api.js
abduegal: agreed, I'll change that

extensions/output/api.js
abduegal: okay I will do that

abduegal:
@storm1er the output extension works the same as the notifiers, exchanges and strategies extensions. For that reason I did not add any manual for it.

With regards to the rest api. The config-sample.js should be self explanatory. I will add the available url and path names in a readme. 

Also, I do not know how I can split this pull request even further, all it contains is REST api support for zenbot.

extensions/output/api.js
abduegal: the random_port function already searches for a free port, so I dont think this is neccesary anymore. The function also tests explicitly if the port is available through a try / catch mechanism

abduegal: image

DeviaVir: Ah, sweet! Thanks for digging that up.

DeviaVir (approved):
:+1: LGTM now. Thanks for this great addition!

albabosh:
Seems that REST API requires two ports. If I put a single port in the config the app crashes a start. Could you give me a working config example ?

storm1er:
@albabosh I don't have any issues with this PR,
Can you explain what's your problem ?

albabosh:
If I enable REST API in conf.js & set c.output.api.port = 20002 it crashes a start (seems that express tries to bind twice to one port). I I set c.output.api.port = 0 zenbot starts well and binds to 2 random ports. I need to specify those two ports in confg (not random)

storm1er:
hmm That's weird ... I didn't need have this problem ... Didn't need to manually set port anyway ^^'
Try to look what they said in review, probably a good start

PR #812, opened 2017-12-07
Full support on Windows and Some fixes for CEXIO

ArnaudRitti:
Removed Talib and Forex dependencies, that provide a complet support under Windows.

For CEXIO

- Increase retry time for CEXIO to avoid API Rate Limit

-  Hide some cosmetic errors


DeviaVir:
I don't think I can accept this PR, gaining Windows support over losing so many strategies and indicators is not worth it IMO.

albabosh:
Could you accept CEX.IO patches only ?

DeviaVir:
@albabosh sure, if you could create a separate PR for that I will look at that.

PR #826, opened 2017-12-09
stddev calculate on period

ghost:
less api calls

### DIFF
--- a/extensions/strategies/stddev/strategy.js
+++ b/extensions/strategies/stddev/strategy.js
@@ -13,8 +13,8 @@ module.exports = function container (get, set, clear) {
       this.option('min_periods', "min_periods", Number, 1250)
     },
     calculate: function (s) {
-    calculated = null
-  },
+      calculated = null
+    },
     onPeriod: function (s, cb) {
       get('lib.ema')(s, 'stddev', s.options.stddev)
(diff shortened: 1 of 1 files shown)


extensions/strategies/stddev/strategy.js
```diff
@@ -13,7 +13,10 @@ module.exports = function container (get, set, clear) {
       this.option('min_periods', "min_periods", Number, 1250)
     },
     calculate: function (s) {
-      get('lib.ema')(s, 'stddev', s.options.stddev)
+    calculated = null
+  },
+    onPeriod: function (s, cb) {
+        get('lib.ema')(s, 'stddev', s.options.stddev)
```
DeviaVir: Odd indenting. Also, what calls the `calculate()` fn now?

JensvdHeydt: the caluculate method get's called with every single trade the bot receives. I don't think that we have to calculate the stddev on every trade but only when it's actually needed: In the onPeriod - call. It's not perfect but it's enough, imho.

ghost: yea ^^

DeviaVir (approved):
:+1:

PR #841, opened 2017-12-12
 Catching more recoverable connection errors in Kraken

MT00x:
Catching more recoverable connection errors in Kraken and fix some debug logs


DeviaVir (approved):
.. damn Kraken.

Issue #800, opened 2017-12-05
Wrong price in Market order 


MT00x:
In live trading I switched to "taker" mode.
Bot made a  MARKET BUY ORDER and the order was filled at higher price than it thought (181.68 vs 180.64€).
The actual buy price was ignored.

See attachment
![bug](https://user-images.githubusercontent.com/10209506/33601537-da64e714-d9ac-11e7-9264-76309ad7eae2.png)

MT00x:
Fixed and merged #847

PR #847, opened 2017-12-12
Fix market order actual price

MT00x:
In market order the actual price (api_order.price) should be very different from trade price
Use actual price if possible.
This should fix issue #800

### DIFF
--- a/lib/engine.js
+++ b/lib/engine.js
@@ -293,6 +293,7 @@ module.exports = function container (get, set, clear) {
               msg('order status: ' + order.status)
               if (api_order.status === 'done') {
                 order.time = new Date(api_order.done_at).getTime()
+                order.price = api_order.price || order.price // Use actual price if possibile. In market order the actual price (api_order.price) should be very different from trade price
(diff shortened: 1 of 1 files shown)


lib/engine.js
```diff
@@ -293,6 +293,7 @@ module.exports = function container (get, set, clear) {
               msg('order status: ' + order.status)
               if (api_order.status === 'done') {
                 order.time = new Date(api_order.done_at).getTime()
+                order.price = api_order.price || order.price // Use actual price if possibile. In market order the actual price (api_order.price) should be very different from trade price
```
DeviaVir: `In market order the actual price (api_order.price) should be very different from trade price` should be or could be?

DeviaVir (approved):

lib/engine.js
MT00x: Could be*, and very often is :)

DeviaVir (approved):
:+1: nice! I missed the typo there myself :)

PR #858, opened 2017-12-13
Add crossover_vwap strategy to readme & genetic backtester.

tiagosiebler:
This opens up this new strategy to genetic backtesting. 

In the process I added a RangeFactor method to the phenotype.js create() function, allowing ranges in factors of x (e.g 1-1000 at a factor of 100). Not 100% sure yet on the RangeFactor implementation, but it seems to work. Setting a min value of 0 also allows the possiblity of a 0 value being returned. E.g: 
`RangeFactor(0, 10000, 10)//0 disables this max cap. Test in increments of 10`


tiagosiebler:
Added some HTML charts as well to demonstrate the effects of the vwap_max parameter in the decision making process.

DeviaVir (approved):
:+1: cool, thanks for contributing!

PR #867, opened 2017-12-14
Created DEMA strategy

benasanders:
It is based off of: https://gekko.wizb.it/docs/strategies/example_strategies.html#DEMA
Very simple, yet effective in some situations.

### DIFF
--- a/extensions/strategies/dema-lessnoise/strategy.js
+++ b/extensions/strategies/dema-lessnoise/strategy.js
@@ -1,0 +1,80 @@
+var z = require('zero-fill')
+  , n = require('numbro')
+
+module.exports = function container (get, set, clear) {
+  return {
+    name: 'dema',
+    description: 'Buy when (short ema > long ema) and sell when (short ema < long ema).',
+
+    getOptions: function () {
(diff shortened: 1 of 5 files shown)


extensions/strategies/dema-lessnoise/strategy.js
```diff
@@ -0,0 +1,83 @@
+var z = require('zero-fill')
+  , n = require('numbro')
+
+module.exports = function container (get, set, clear) {
+  return {
+    name: 'dema',
+    description: 'Buy when (short ema > long ema) and sell when (short ema < long ema).',
```
DeviaVir, line 7: How is it "lessnoise"? Also I think you need to update the name here?

yura505: Just FYI, using VMA as long MA will filter noise better, because of it adjusts volatility. Try EMA10/VMA10, on 15 or 30 min intervals. Of cause it works only in strong trends.

DeviaVir:
This issue was moved to highvelocityspace/zenbot#81

DeviaVir (approved):
I'll approve and merge since more strats == better

PR #869, opened 2017-12-14
Less preroll this way

ghost:
(I have a fix for gdax, basically, workaround to just supress the errors.....)


ghost:
I need to update stddev strategy........
As well, ty for not pulling the gdax code but I think this is still a viable workaround for gdax atm:


```
  function retry (method, args, err) {
    if (method !== 'getTrades') {
    Timeout(function () {
      exchange[method].apply(exchange, args)
    }, 10000)
  }
```

PR #883, opened 2017-12-15
[CHG] SIM: use human datetime format for start/end parameter

tfossoul:
You can use "YYYYMMDDhhmm" like "201712010800" instead of timestamp format.


DeviaVir (approved):
:+1:

PR #884, opened 2017-12-15
make sure that a wrong API key, will yield an exception to save some …

sassman:
Had a wrong api-secret and it crashed with a number format issue because s.balance was not properly initialized. 

save some time of investigating whats wrong.
can be reproduced with a correct api key but incorrect secret

### DIFF
--- a/commands/trade.js
+++ b/commands/trade.js
@@ -312,7 +312,6 @@ module.exports = function container (get, set, clear) {
         var db_cursor, trade_cursor
         var query_start = tb().resize(so.period).subtract(so.min_periods * 2).toMilliseconds()
         var days = Math.ceil((new Date().getTime() - query_start) / 86400000)
-        var trades_per_min = 0
         var session = null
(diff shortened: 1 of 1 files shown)


commands/trade.js
```diff
@@ -467,6 +466,10 @@ module.exports = function container (get, set, clear) {
         function forwardScan () {
           function saveSession () {
             engine.syncBalance(function (err) {
+              if (!err && s.balance.asset === undefined) {
```
DeviaVir, line 469: My worry is that some exchanges might return undefined for empty balances, and we would tell the user their API key is not working even though it is. My worry is just an assumption though so if nobody shares this worry I'm fine with merging this.

sassman: I thought so as well, but if asset or currency is not defined in that balance object, then a lot of other code will crash for sure. This way or that way it dies. :(

DeviaVir (approved):
:+1:

PR #888, opened 2017-12-16
HTML report graphs use green arrows when buying

Wood-Chopper:
It's easier to read the graph, especially for graphs with a lot of buy/sell.


DeviaVir (approved):
:+1: cool, thanks for contributing!

PR #892, opened 2017-12-16
Enable short position to lose even more money !!! + Ichimoku strategy

slashafk:
Hello, 
_(This is the first time i'm doing a PR on github so i'm not sure about how it actually works)._

This patch is about adding **short and close position** to earn money when markets are going down. Those order in general cost a taker fee or have specific fee (depends on exchanges if exchange enable it). 
**So far it only support position on cex.io (the only exchange i'm using for test).**

I'm not sure about how the order system works ... lots of copy/paste to have something working ...
the function **executeOrder** is doing ~200 lines of code and is really hards to play with.

But it works on simulation (not completly sure about the calculation of balance on sim, lots of balance calculation everywhere in the engine.js :/) and short /close are executed on sim.

_A refactoring of that part could be helpful._

---

I also added a **new (actually VERY old) strategy** : [Ichimoku Kinko Hyo](https://en.wikipedia.org/wiki/Ichimoku_Kink%C5%8D_Hy%C5%8D).

This strategy is supposed to give good result on high volatility markets going up or down (supports short positions very well), it's also able to detect sideway markets but it's very difficult to do it programatically so this version doesn't detect those (could be an interesting improvment to use limit inside the support/resistance).

_This strategy gives some good results on sim ... but as always fails on live :/._
The Ichimoku strategy has **6 signals** (2 strongs, 2 weaks and 2 mediums) i've only implemented the 2 strongs (the buy and short with their close). The weaks and mediums signals add lots of errors so they should probably used with others indicators (RSI, CCI, ...) but no documentation about that.

This strategy has been originally developped for a 4h period (got some simulation with a 0% error on a 4h period), but it can work on other periods (with more errors ;)).

---
**I'm not responsible for any loss using this strategy ;)**


extensions/strategies/ichi/strategy.js
```diff
@@ -0,0 +1,143 @@
+var z = require('zero-fill')
+, n = require('numbro')
+
+/**
+ * Calculate if the Kumo can be reversed (sideway market or kumo twist)
+ * */
+function isReversable(s){
+	var kumo_size = s.period.close*0.005;
+	//last_kumo = Math.abs(s.period.ssa-s.period.ssb)<kumo_size
```
evseevnn-zz: Please remove code what your not use.

slashafk:
Well, I just found an issue where the bot can't close a position .... the order system is a serious nightmare :/

extensions/strategies/ichi/strategy.js
slashafk: this code will/should be use, i'm trying to implement the kumo analysis (a bit harder than what I originally thought).

extensions/strategies/ichi/strategy.js
```diff
@@ -0,0 +1,143 @@
...
+
+    /*TODO : WTF ???
+    */
+  calculate: function (s) {
+    //some additionnal indicator useful one day ;)
+    get('lib.ema')(s, 'trend_ema', s.options.trend_ema)
+    get('lib.cci')(s, 'cci', s.options.cci_periods,s.options.cci_constant)
+    get('lib.rsi')(s, 'rsi', s.options.oversold_rsi_periods)
+    //Ichimoku
+    get('lib.midprice')(s, 'ts', 9)
```
nedievas: 20

extensions/strategies/ichi/strategy.js
```diff
@@ -0,0 +1,143 @@
...
+    get('lib.midprice')(s, 'ks', 26)
```
nedievas: 30

DeviaVir:
This issue was moved to highvelocityspace/zenbot#47

slashafk:
I'll split the strategy and the position (which require a lot of checks and improvments for the sim)

PR #894, opened 2017-12-16
Improve XMPP reliability

defkev:
The XMPP notifier is awfully unreliable, at least for me, due to the connect/disconnect at every notification, especially if the bot is happily doing what he is supposed to do - trade.

At some point my XMPP server will start rate-limiting the connection causing a whole lot of notifications to never get delivered.

This will make the notifier act more like a real XMPP client by only connecting once, keeping the connection open and reconnecting if necessary.


DeviaVir (approved):
:+1:

PR #896, opened 2017-12-16
Slow Stochastic Oscillator

yura505:
Buy when (K > D) && (D < 30)
Sell when (K < D) && (D > 70)
Low profit, but low error rate. This oscillator works good in trending range, but can be used for trends too.


DeviaVir (approved):
:+1: thanks for contributing!

PR #897, opened 2017-12-16
Adding stackoverflow link to README file.

dshetyo:
Added stackoveflow link README file. 
Link is https://stackoverflow.com/questions/tagged/zenbot


DeviaVir (approved):
I started using SO for this only recently, there has been some backlash from SO users about if this is the right thing to do, meanwhile I'll merge this and we can switch to something else later if needed.

Issue #901, opened 2017-12-17
[Bitstamp] Order completion not recognised by zenbot


tiagosiebler:
### System information
- **Have I written custom code (as opposed to using zenbot vanilla)**: vanilla zenbot
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: linux
- **Zenbot version**: latest
- **NodeJS version**: 8
- **Python version**: n/a
- **Exact command to reproduce**: trade live with bitstamp

### Describe the problem
Seems zenbot can't understand the api response when an order completes via bitstamp. Currently testing a fix, just logging this to keep track.

### Source code / logs
```
2017-12-17 00:24:28 - funds on hold after cancel, waiting 5s

 ???? Bitstamp.getOrder() err, body:  null { status: 'Finished',
  transactions: 
   [ { fee: '0.45000000',
       price: '0.62000000',
       datetime: '2017-12-16 22:06:58',
       tid: 33971784,
       type: 2,
       xrp: 'xxxx',
       eur: 'xxxx' } ] }
```

PR #903, opened 2017-12-17
[Bitstamp] Fix order completion detection & balance reading during open order.

tiagosiebler:
The key fix is for issue #901. Zenbot places an order on bitstamp and doesn't realise it's completed when the order status comes back as "Finished". I've added some logic to change it to "done" when that happens.

When an order completed, the message from zenbot couldn't read the date returned from the API:
`sell order completed at Invalid date:`

Added a bit of logic to map datetime to date, so the engine.js logic can read it properly. Lastly, while an order is open/placed with bitstamp, the current logic read the current balance as 0 and the balance on hold was hardcoded as 0. I've mapped that correctly to x_balance and x_reserved from the bitstamp API response.

### DIFF
--- a/extensions/exchanges/bitstamp/exchange.js
+++ b/extensions/exchanges/bitstamp/exchange.js
@@ -242,14 +242,21 @@ module.exports = function container (get, set, clear) {
         if (body.status === 'error') {
 	        return retry('getBalance', args)
         }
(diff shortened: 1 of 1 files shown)


DeviaVir (approved):
:+1: let's see it in the wild

PR #906, opened 2017-12-17
replace normalize selector paradigm with hopefully cleaner objectify selector paradigm

haxwell:
Hello, this is a first pull request for me.. Please be gentle. :) I have been trying to read and understand the code, and this is a small change that I think improves the readability of the code. Specifically, where there was code to slice the 'selector' string to get exchange, product, etc., now that code is abstracted, and we use object properties to get those 'selector' attributes.


DeviaVir (approved):
:+1: thanks for contributing this, I agree in thinking that this will make it easier for beginners to jump in.

DeviaVir (commented):
I merged latest master into this branch, would you like to check it one more time before I merge?

haxwell:
Yes, I will review the changes today, and comment again.

haxwell:
There is some minor cleanup that could be done there.. for instance, balance.js uses `s.selector` while the other commands generally use `so.selector`... but it was that way before this change, and still works. 

So, as the mocha unit testing refactoring happens, I think some issues like the above will clear up automatically.

Ship it!

PR #907, opened 2017-12-17
Expansion on PR #854, missed this hardcoded section in the auto-dump (shift + D).

tiagosiebler:
Expansion on PR #854, missed this hardcoded section in the auto-dump (shift + D).

> Instead of hardcoding which values are included in the export, include all values in the export. Allows for further enhancement of the graphs. #854


DeviaVir (approved):
:+1: cool, thanks for contributing

PR #908, opened 2017-12-17
Add exchange-specific tips to help others avoid similar headaches.

tiagosiebler:
We should start documenting settings we find that we need to change from defaults, to have Zenbot work better with specific exchanges. I've added comments based on my experiences with Bitstamp & Kraken. These were barely usable at first, but seem to be pretty reliable now.


DeviaVir (approved):
Yes, a million times yes, thanks for writing this up.

Issue #769, opened 2017-11-29
Custom config file not overwritte  mongo.db param ?


TheRealBobi:
I give a custom_conf.js in command line  like : 

`sim --conf JDJiJDEyJDEvSHd3SnJVMmNuWno5a1RxbUZ0WU9yVTlQZWVjTW1pU2NCUC53RkhVNy5QRkQzZnI1VHU2_conf.js  gdax.LTC-EUR    --srsi_macd   --order_type maker   --reset_profit yes`

Have set mongodb params in this custom file but zenbot using default db params in conf.js. 

How can i fix it ?

haxwell:
This should no longer be an issue with PR #912.

PR #912, opened 2017-12-18
Issue #769, added code which attempts to load the configuration file …

haxwell:
Issue #769, added code which attempts to load the configuration file passed in on the command line. If not found, tries ./conf.js. Previously, it was only trying ./conf.js.


DeviaVir (approved):
:+1: that's great! Thanks for contributing

PR #920, opened 2017-12-18
Add telegram notifier

Want100Cookies:
I was missing the Telegram notification so I added it. It uses the well known [yagop/node-telegram-bot-api](https://github.com/yagop/node-telegram-bot-api) package and a bot token and chat id should be provided through the config file.

### DIFF
--- a/conf-sample.js
+++ b/conf-sample.js
@@ -214,6 +214,13 @@ c.notifiers.textbelt.phone = '3121234567'
 c.notifiers.textbelt.key = 'textbelt'
 // end textbelt configs
 
+// telegram configs
+c.notifiers.telegram = {};
+c.notifiers.telegram.on = false // false telegram disabled; true telegram enabled (key should be correct)
+c.notifiers.telegram.bot_token = 'YOUR-BOT-TOKEN'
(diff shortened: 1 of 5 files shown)


conf-sample.js
```diff
@@ -214,6 +214,13 @@ c.notifiers.textbelt.phone = '3121234567'
 c.notifiers.textbelt.key = 'textbelt'
 // end textbelt configs
 
+// telegram configs
+c.notifiers.telegram = {};
+c.notifiers.telegram.on = false // false telegram disabled; true telegram enabled (key should be correct)
+c.notifiers.telegram.bot_token = 'YOUR-BOT-TOKEN'
+c.notifiers.telegram.chat_id = '8942311' // the id of the chat the messages should be send in
```
DeviaVir: should this be empty?

DeviaVir (approved):

conf-sample.js
Want100Cookies: Yeah, it should be. I will make it _YOUR-CHAT-ID_

DeviaVir (approved):
:+1:

PR #926, opened 2017-12-19
rename so period to so period size

haxwell:
This leaves the command line parameter --period in place, but deals with the value for it within the code as 'periodSize'. This name more accurately describes what is in the variable. This is especially true in light of the s.period object created in engine.js. That object describes high and low values over a period of time.

I'm not stuck on the name periodSize.. periodLength I think is appropriate, too. The phrase 'period length' is used in the description for the command line variable 'period'.

### DIFF
--- a/commands/backfill.js
+++ b/commands/backfill.js
@@ -154,7 +154,7 @@ module.exports = function container (get, set, clear) {
                     if (err) throw err
                     trade_counter += trades.length
                     day_trade_counter += trades.length
-                    var current_days_left = Math.ceil((mode === 'backward' ? marker.oldest_time - target_time : target_time - marker.newest_time) / 86400000)
+                    var current_days_left = 1 + (mode === 'backward' ? tb(marker.oldest_time - target_time).resize('1d').value : tb(target_time - marker.newest_time).resize('1d').value)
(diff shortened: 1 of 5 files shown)


haxwell:
Not sure why that a393830 commit appears.... :/

DeviaVir (commented):
I'm not yet convinced what's wrong with just letting it stay `period`. This seems to add more overhead than that it gives us

haxwell:
I think it would appear less overhead, ironically, if the PR was taken a little further, actually.. This is an effort to keep two concepts in the code clear and distinct. We use the term 'period' for two different concepts. First, to describe the length of a candlestick, and second to refer to an object which keeps the metadata about a candlestick.

When I say take it further, I mean that perhaps the command line variable should be changed to --period(size|length) _or_ --candlestick(size|length), as well. The names of those command line options become automatic variable names in the code. 

So, even though the user never deals with the metadata object called 'period', they do create another variable called 'period', that the developers deal with. And it has a different meaning than another object that the developers have given the same name. 

Just trying to lower the learning curve, for the next guy. :100:

haxwell:
Fat finger closing.

DeviaVir (approved):

Issue #938, opened 2017-12-19
Error: invalid bucket size spec: undefined


efams:
Hi,
after last git pull i get this error when i try to start a simulation.
Lubuntu: 17.10

Error: invalid bucket size spec: undefined
    at BucketSize.parse (/home/enri/gdax/zenbot/node_modules/timebucket/lib/bucket-size.js:15:21)
    at new BucketSize (/home/enri/gdax/zenbot/node_modules/timebucket/lib/bucket-size.js:5:21)
    at Bucket.resize (/home/enri/gdax/zenbot/node_modules/timebucket/lib/bucket.js:30:14)
    at Command.<anonymous> (/home/enri/gdax/zenbot/commands/sim.js:84:51)
    at Command.listener (/home/enri/gdax/zenbot/node_modules/commander/index.js:315:8)
    at emitTwo (events.js:126:13)
    at Command.emit (events.js:214:7)
    at Command.parseArgs (/home/enri/gdax/zenbot/node_modules/commander/index.js:653:12)
    at Command.parse (/home/enri/gdax/zenbot/node_modules/commander/index.js:475:21)
    at /home/enri/gdax/zenbot/zenbot.js:39:11
    at /home/enri/gdax/zenbot/boot.js:24:7
    at f (/home/enri/gdax/zenbot/node_modules/once/once.js:25:25)
    at Glob.<anonymous> (/home/enri/gdax/zenbot/node_modules/glob/glob.js:151:7)
    at emitOne (events.js:116:13)
    at Glob.emit (events.js:211:7)
    at Glob._finish (/home/enri/gdax/zenbot/node_modules/glob/glob.js:197:8)
    at done (/home/enri/gdax/zenbot/node_modules/glob/glob.js:182:14)
    at Glob._processGlobStar2 (/home/enri/gdax/zenbot/node_modules/glob/glob.js:637:12)
    at /home/enri/gdax/zenbot/node_modules/glob/glob.js:626:10
    at RES (/home/enri/gdax/zenbot/node_modules/inflight/inflight.js:31:16)
    at f (/home/enri/gdax/zenbot/node_modules/once/once.js:25:25)
    at lstatcb_ (/home/enri/gdax/zenbot/node_modules/glob/glob.js:519:7)

Thx
Enri

haxwell:
I'll take this.. I think it is somehow related to my PR #926.

haxwell:
See PR #940 .

canselcik:
[...]

PR #940, opened 2017-12-20
#938, Fixing Error: invalid bucket size spec: undefined

haxwell:
In addition to the options on a command, there are options on 
strategies. There is a callback function in engine.js which sets the
options as defined in a strategy. The strategy objects were not
updated to have a periodSize option, only a period option. So,
the periodSize attribute was not being set. This caused the
execption in this bug.

This commit adds a second option, named 'periodLength'. The name
'periodLength' was chosen, rather than the previous 'periodSize' for
consistency; the description for the --period option already said
'period length', so it made sense that the additional variable be named
that as well.

Each reference to *.periodSize has been changed to *.periodLength.

This commit continues PR #926.

### DIFF
--- a/commands/sim.js
+++ b/commands/sim.js
@@ -49,7 +49,7 @@ module.exports = function container (get, set, clear) {
           }
         })
 
-        so.periodSize = so.period
+        so.periodLength = so.period
 
         if (so.start) {
           so.start = moment(so.start, "YYYYMMDDhhmm").valueOf()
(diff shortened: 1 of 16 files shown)


DeviaVir (approved):
:+1:

dsrikanth:
@haxwell Thanks for doing this! Do you see changes needed on the backtester scripts?

efams:
Thx, now it's all ok ;)

efams:
If i use --strategy stddev returns me the same error "Error: invalid bucket size spec: undefined"

haxwell:
@dsrikanth Thank you for mentioning the backtesters; I didn't know they were there. 

@efams A likely excuse, I know, but the filter I was using to find where to replace strategy option definitions for --period, missed stddev, and two other strategies. I did a more manual search, to get all the occurrences.

I've submitted PR #960 and #961 to address both the above points.

Thanks for your patience, everyone.

PR #945, opened 2017-12-20
CEXIO api update to version 1.0.8 - solves Nonce increment error

nedievas:
`npm install` after `git pull` and generate new API keys on exchange!


DeviaVir (approved):
:+1:

PR #953, opened 2017-12-20
GDAX getQuote: return an error if market is empty

fuzzyTew:
This makes the failed balance output for bch-usd during this market freeze make more sense.


DeviaVir (approved):
:+1:

PR #967, opened 2017-12-21
CEXIO use "my" fees

defkev:
Get (and use) personalized fees on CEXIO as returned by the API at https://cex.io/api/get_myfee for live and paper trading.

The fees get recalculated (by CEX) and refreshed (by us) every day at midnight (GMT)

The timer uses UTC so this works regardless of the bots local timezone.

### DIFF
--- a/lib/engine.js
+++ b/lib/engine.js
@@ -30,6 +30,9 @@ module.exports = function container (get, set, clear) {
       console.error('error: could not find product "' + s.product_id + '"')
       process.exit(1)
     }
+    if ((so.mode === 'live' || so.mode === 'paper') && selector_parts[0] === 'cexio') {
+      s.exchange.setFees({asset: s.asset, currency: s.currency})
+    }
     if (so.mode === 'sim' || so.mode === 'paper') {
       s.balance = {asset: so.asset_capital, currency: so.currency_capital}
     }
--- a/lib/engine.js
+++ b/lib/engine.js
@@ -30,7 +30,7 @@ module.exports = function container (get, set, clear) {
       console.error('error: could not find product "' + s.product_id + '"')
       process.exit(1)
     }
-    if ((so.mode === 'live' || so.mode === 'paper') && selector_parts[0] === 'cexio') {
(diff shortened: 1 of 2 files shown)


nedievas:
Good! I was planning to add dynamic fees to Bitfinex and Cexio exchanges. This looks like even better approach. Let's test in the wild?

lib/engine.js
```diff
@@ -30,6 +30,9 @@ module.exports = function container (get, set, clear) {
       console.error('error: could not find product "' + s.product_id + '"')
       process.exit(1)
     }
+    if ((so.mode === 'live' || so.mode === 'paper') && selector_parts[0] === 'cexio') {
+      s.exchange.setFees({asset: s.asset, currency: s.currency})
+    }
```
DeviaVir: Is there any way to prevent this from ending up as an exchange-specific hack in `lib/engine.js`?

defkev: Was initially considering adding the setFees: function() to every exchange (as a pseudo call doing nothing for now except for CEX) but didn't want to change to many files.

But if this gets adopted for other exchanges i don't see a reason to not do it in a cleaner/less hackish/none exchange-specific way.

Your call :)

defkev:
@nedievas 
I am already running this on one of my CEX bots for +24 hours and so far no problems.

As long as the exchange and API-module supports it (like yours does) this should be fairly easy to adopt for other exchanges (change the pair format, JSON fields and adjust the time the update should occur)

The only odd thing i noticed about CEX is that the API returns four fees in total (taker buy/sell and maker buy/sell) will the front page only lists maker/taker fees.

The buy/sell fees are the same (at least for me) so this uses the median of each.

lib/engine.js
defkev: On a second thought i could add a new exchange property (called dynamicFees or whatever) and check for it in engine.js

Let me update the PR

defkev: Done

DeviaVir:
This issue was moved to highvelocityspace/zenbot#18

DeviaVir (approved):
:+1: awesome work, and thanks for refactoring it!

PR #969, opened 2017-12-21
Added support for docker-compose 3 and fixed dependencies between containers.

alincalinciuc:
Added support for docker-compose 3 and fixed dependencies between containers.


DeviaVir:
This issue was moved to highvelocityspace/zenbot#17

DeviaVir (approved):
:+1: thanks for contributing, let's see it in the wild

PR #971, opened 2017-12-21
Don't generate file if filename is none

krystophv:
When filename is set to none, we still go through the efforts of making the file, just not writing it. It'd be better if we didn't generate any of those contents at all.


DeviaVir (approved):
:+1:

PR #974, opened 2017-12-22
Jasmine refactor/backfill command/creating collections service2

haxwell:
So, this commit shows where I'm going with unit tests. There is a lot of duplicate functionality across the commands, strategies, and exchanges. I want to DRY that up, using one service function for each place we currently call that duplicate functionality.

To that end, I've created collection-service.js, and called it from the command backfill.js. Also, to the point of this whole effort, I created a unit test for the collection service.


JensvdHeydt:
You're completely right. There is duplicate code all over the place and it's very hard to refactor to achieve testability. Also having functions declared inside of functions is bad. It's very hard to understand the code flow in the project's current state.

lib/_codemap.js
```diff
@@ -20,5 +20,7 @@ module.exports = {
   'lrc': require('./lrc'),
   'adx': require('./adx'),
   'vwap': require('./vwap'),
-  'slow_stochastic': require('./slow_stochastic')
+  'slow_stochastic': require('./slow_stochastic'),
+
+  'collection-service': require('./services/collection-service')
```
JensvdHeydt: I think, "/services/collection-service'" should be named "/services/collection". No need to duplicate the "service". What do you think?

JensvdHeydt:
@haxwell @DeviaVir I'd like to have automatic unit testing and codestyle checks using Travis CI. Do we have enough access rights to change the project settings?

lib/_codemap.js
haxwell: True, there's no _need_, still I think it adds more than it hurts.

DeviaVir:
@JensvdHeydt no, I do not. I've been thinking about moving it to a fork I control, but that's the "nuclear" option

tiagosiebler:
@DeviaVir if we can't get hold of @carlos8f and he's the only one able to give these privileges, I think it's an option worth considering. Hard fork with credit to original author.

DeviaVir:
@tiagosiebler reached out to github support, they've been helpful in the past, maybe they have a solution for this too.

DeviaVir:
They've given me a new API endpoint that will allow me to migrate all issues. I've created https://github.com/carlos8f/zenbot/issues/1010 to track and get more opinions.

DeviaVir (approved):

PR #976, opened 2017-12-23
Tweaked XRP increments for Bitstamp & fixed invalid date messages

tiagosiebler:
- XRP increments were too big, reduced those. 
- When orders completed on bitstamp, the completion logic couldn't get the datetime properly. This fixed it for me, so I'm sharing it here.

### DIFF
--- a/extensions/exchanges/bitstamp/products.json
+++ b/extensions/exchanges/bitstamp/products.json
@@ -57,18 +57,18 @@
     "id": "XRPUSD",
     "asset": "XRP",
     "currency": "USD",
-    "min_size": "0.01",
+    "min_size": "5",
     "max_size": "1000000",
-    "increment": "0.01",
+    "increment": "0.00001",
     "label": "XRP/USD"
   },
   {
     "id": "XRPEUR",
     "asset": "XRP",
     "currency": "EUR",
-    "min_size": "0.01",
+    "min_size": "5",
     "max_size": "1000000",
-    "increment": "0.01",
+    "increment": "0.00001",
     "label": "XRP/EUR"
   },
   {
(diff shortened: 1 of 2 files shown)


extensions/exchanges/bitstamp/products.json
```diff
@@ -57,18 +57,18 @@
     "id": "XRPUSD",
     "asset": "XRP",
     "currency": "USD",
-    "min_size": "0.01",
+    "min_size": "5",
```
haxwell, line 60: I think min_size refers to the minimum quantity of the asset that you can order at one time. If someone else can confirm that, then my next question is, wouldn't a min_size of 5 mean you had to order a minimum of 5 at once? Possible, but doesn't sound right.

tiagosiebler: Yes, your statement is correct. I've made this change because bitstamp requires a minimum trade amount of 5 euros - see end of page:
https://www.bitstamp.net/fee_schedule/

It's somewhat hard to set for a currency that's fluctuating at prices below 1 euro, but this alleviates the issue that zenbot currently tries to trade at smaller amounts, repeatedly. I don't really use the USD amount, but thought increasing the bar on that might avoid others some pain.

tiagosiebler: Quote for convenience
>The minimum allowable trade is 5 Euro for Euro denominated trading pairs. The minimum allowable trade is 0.001 BTC for BTC denominated trading pairs. The value of minimum trades in other trading pairs will be denominated in the fiat currency of the particular trading pair and limited to 5 whole units of that currency (example: $5, £5, etc.).

DeviaVir (approved):
:+1:

PR #977, opened 2017-12-23
Include trendline strategy for genetic backtesting

yang2lalang:
This change enables us to  Backtest with trendline Strategy using the genetic backtester


DeviaVir (approved):
:+1:

PR #982, opened 2017-12-24
Fixed bug where the "stop loss" protection was not triggering

acecilia:
I was simulating. My configuration specifies that:

    c.sell_stop_pct = 1
    c.buy_stop_pct = 1

While simulating I saw this:

captura de pantalla 2017-12-25 a las 2 26 05

This makes no sense, because the stop loss of 1% (c.sell_stop_pct = 1) should have triggered. This is a bug (I think).

I debugged it, and found that:
- This happens when the buy order is triggered by a "buy stop" protection (c.buy_stop_pct = 1).
- In such a case, when the "buy stop" protection triggers, `s.acted_on_stop` is set to true (line 178). As a consequence, `s.sell_stop` does not get assigned a value (line 601).
- As a result, the stop loss (c.sell_stop_pct = 1) is disabled (in order for the stop loss to work, `s.sell_stop` must have a value (line 157)).

Happy holidays ;P

### DIFF
--- a/lib/engine.js
+++ b/lib/engine.js
@@ -598,7 +598,7 @@ module.exports = function container (get, set, clear) {
           delete s.buy_order
           delete s.buy_stop
           delete s.sell_stop
-          if (!s.acted_on_stop && so.sell_stop_pct) {
+          if (so.sell_stop_pct) {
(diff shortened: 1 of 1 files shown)


DeviaVir (approved):
:+1: thanks for debugging this

PR #987, opened 2017-12-26
Chaikin Money Flow oscillator

yura505:
Chaikin Money Flow should not be used as a stand-alone indicator, but it's very useful for trend confirmation by volume. Positive CMF would confirm an uptrend, but negative CMF would call into question the strength behind an uptrend. The reverse holds true for downtrends. I use it with length 20 and threshold 0.05


DeviaVir (approved):
:+1: thanks for contributing

PR #990, opened 2017-12-26
fix exchange response key

leonzinger:
The key in the response from Kraken had X as the prefix for each currency. So i have fixed that in code to remove the error when tried to perform a
`zenbot backfill kraken.LTC-XBT --days 30`
command


DeviaVir (approved):
Interesting, thanks for contributing!

PR #995, opened 2017-12-26
Small fixes for drops in objectify-selector changes

fuzzyTew:
#906 lost some codelines that broke the CLI commands balance, buy, and sell.  This patch adds them back so these commands will work again.


haxwell (approved):
Thanks!

DeviaVir (approved):
:+1:

PR #1000, opened 2017-12-27
Revert #990

tiagosiebler:
Can we revert PR #990 ?

This breaks API calls for fiat currencies, see comments here: https://github.com/carlos8f/zenbot/commit/70e7bbc93224a0bada2beff90bfe1b5db2d88a1f


yura505:
I think better to fix the issue instead of revert all enhancements

tiagosiebler:
@yura505 this should only revert the change from PR #990:
https://github.com/carlos8f/zenbot/pull/1000/files#diff-ecb8cbfdc5b6e4b637f435eba0f4cff2
Made this PR using the 'revert' button here on github. 

Will it revert the other enhancements as well? The files changed only shows the kraken exchange.js file

tiagosiebler:
Submitted a cleaner PR #1009.

PR #1009, opened 2017-12-28
Revert #990. Cannot assume every kraken currency has the same prefix

tiagosiebler:
Can we revert PR #990 ?

This breaks API calls for fiat currencies, see comments here: https://github.com/carlos8f/zenbot/commit/70e7bbc93224a0bada2beff90bfe1b5db2d88a1f


DeviaVir (approved):

DeviaVir:
This issue was moved to highvelocityspace/zenbot#3

PR #850, opened 2017-12-12
Visual dashboard for REST API

ArnaudRitti:
Dashboard for zenbot

It work on tablet, desktop and mobile.
On mobile you can add it to home screen and it works like a app.

screenshot_20171212-231925

fireshot 1

PR #1013, opened 2017-12-28
REST API UI v1.1

kernelsndrs:
This is mostly a transfer of code from PR #850, with a refactor from bower to webpack and few small changes related to bootstrap4.

**TODO**
move to partials, integrate ajax updating
integrate --api flag which sets auto stat dumping and prevents toggle
explore best representation for multiple running zenbot instances


kernelsndrs:
welp rebase gone wrong, will resubmit this once its cleaned up.

PR #1029, opened 2018-01-01
For #1002 - ensure start/end are passed along to future calls

glennfu:
My PR for #1002 missed a spot. Future generation calls would get a start/end param that had been modified by `sim.js`, and run against a different time. This fixes it (I hope).


DeviaVir (approved):
:+1:

PR #1032, opened 2018-01-02
For #1002 - while start was fixed, 'end' was still not passing right

glennfu:
Basically a typo fix:
```
  let start = parseInt(params.start);
  let end = parseInt(params.start);
```

Becomes:
```
  let start = parseInt(params.start);
  let end = parseInt(params.end);
```

I was happy to have finally fixed the start param not passing and overlooked that the end param was still not right.


DeviaVir (approved):
:+1: derp I also missed it, good catch

PR #1033, opened 2018-01-02
darwin.js output

shr00mie:
xlync noticed that column A was being output as json instead of normalized.

- changed first field to selector.normalized
- column A was also being used as input for module.selector output in column O, which was populating ""selector"": field with json column A contents instead of the selector.
- this fix also makes the html json parser work and include the selector properly without the [ object Objects ] field.


evseevnn:
@shr00mie Thank for that!

PR #1035, opened 2018-01-03
rename periodLength to period_length

FunFR:
To have the consistency of parameter names, it would be better to rename perdioLength to period_length


DeviaVir (approved):
:+1: thanks for contributing

PR #1043, opened 2018-01-04
Fix bug where fitness function assigned higher fitness to worse results.

ricktonoli:
The fitness calculation does not appear to work correctly when win < loss AND buyvsHold value is negative. It results in a higher fitness value being assigned to worse results. In generation runs where ONLY negative results are returned, each subsequent generation gets progressively worse, not better. To test this I ran trend_ema, 60 days on cexio.ETH_USD and the results now assign the best fitness to the best result correctly, where it was incorrect before.

### DIFF
--- a/scripts/genetic_backtester/phenotype.js
+++ b/scripts/genetic_backtester/phenotype.js
@@ -68,7 +68,8 @@ module.exports = {
     if (typeof phenotype.sim === 'undefined') return 0;
     
     var vsBuyHoldRate = (phenotype.sim.vsBuyHold / 50);
-    var wlRatioRate = 1.0 / (1.0 + Math.pow(2.71828, -(phenotype.sim.wins - phenotype.sim.losses)));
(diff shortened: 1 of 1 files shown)


DeviaVir (approved):
good catch :+1:

scripts/genetic_backtester/phenotype.js
```diff
@@ -68,7 +68,8 @@ module.exports = {
     if (typeof phenotype.sim === 'undefined') return 0;
     
     var vsBuyHoldRate = (phenotype.sim.vsBuyHold / 50);
-    var wlRatioRate = 1.0 / (1.0 + Math.pow(2.71828, -(phenotype.sim.wins - phenotype.sim.losses)));
+    var wlRatio = phenotype.sim.wins - phenotype.sim.losses
```
SlackerATX, line 71: missing ';'?

DarkShimy00:
Why not use `Math.abs` for clarity?

PR #1048, opened 2018-01-04
bring the recent commits to master down into unstable

kernelsndrs:
bring the recent commits to master down into unstable


DeviaVir:
Thanks!

PR #1051, opened 2018-01-05
Added --buy_max_amt option to allow for multiple bots on same exchange

T-rav:
This pull request is for issue https://github.com/DeviaVir/zenbot/issues/174 

If --buy_max_amt is set the bot will trade up to this amount of currency provided it is less than the --buy_pct amount.

If using 1 btc and --buy_pct is set to 50 and --buy_max_amt is set to 0.25, the bot will only trade up to 0.25 btc given it is the lesser of the two amounts.

If using 1 btc and --buy_pct is set to 50 and --buy_max_amt is set to 0.75, the bot will trade 0.5 btc or 50% given it is lesser of the two amounts.

This option is only for live trading. And of course the stats each bot generates will be off when running multiple bots on the same exchange.


DeviaVir (approved):
:+1: this might be the best PR in a while, tested and everything. Thank you

firepol:
I just tested this on paper, I did set an initial balance to 1 BTC, buy_pct set to 10%, --buy_max_amt 0.0025

The simulator just did 2 buy orders of 0.09 BTC and 0.08 BTC. I was expecting to see 0.0025 orders instead.

@T-rav Can you confirm that (as you wrote) this works only in the live trading (but not in the simulator)?
If so, would be nice to make this work also in the simulator... thank you.

Haehnchen:
> This option is only for live trading

@firepol this is only for live trading, but i can confrm that working. This is how i am simulating it for now, not nice but better then nothing:

`zenbot.sh trade binance.XRP-BTC --paper --period=4m --buy_max_amt=0.00326927 --currency_capital=0.00326927 --reset_profit`

firepol:
I just submitted a pull request to make this work also on paper mode, it was merged just a few minutes ago: https://github.com/DeviaVir/zenbot/pull/1099

@Haehnchen what does the --reset_profit option do? I thought that every time I kill zenbot and run it again, it resets the bot, isn't it the case?

PR #1052, opened 2018-01-05
Indicator SRSI

Shawn8901:
Since SRSI accepts a custom key the calucation of StochRSI_D should respect this custom key.
Also i have adjusted the console output to fit the data rows.
Feel free to accept or decline this.

If i should separate both requests let me know.

### DIFF
--- a/lib/engine.js
+++ b/lib/engine.js
@@ -698,7 +698,7 @@ module.exports = function container (get, set, clear) {
         process.stdout.write(z(8, pct(diff), ' ')[diff >= 0 ? 'green' : 'red'])
       }
       else {
-        process.stdout.write(z(8, '', ' '))
+        process.stdout.write(z(9, '', ' '))
       }
       var volume_display = s.period.volume > 99999 ? abbreviate(s.period.volume, 2) : n(s.period.volume).format('0')
       volume_display = z(8, volume_display, ' ')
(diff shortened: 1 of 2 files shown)


yura505:
@Shawn8901 Could you also check that SRSI calculated correctly? Recently I have seen values over 100 there, but it should be in 0..100 range. Thanks.

Shawn8901:
i also seen it, im working on a custom strategie, which uses the SRSI as indicator. The SRSI also goes below 0.
I am confused of all indicators which i use.
If i am checking those with the index listed in www.tradingview.com most of them do not really fit with the calculated of zenbot (but thats another point).

yura505:
@Shawn8901 It because some if indicators are looking back for all time (all known periods), tradingview is looking since BTC was added (years ago), but in zenbot you specifying in in 'min_period' option. Another point is some of indicators ends at last closed period (s.lookback[0]), but not current unclosed period (s.period), what is tradingview does. Other words for some of indicator you see values of previous period and it should be fine if current period is not closed.

yura505:
@Shawn8901 I think if add "RSI.push(s.period.rsi)" between lines 6 and 7 the issue that I wrote should away. Author of srsi.js forgot add current period to array I think.

Shawn8901:
Isn't the parameter k then to high? has it to be reduced? Adding current RSI also was my first try (with better result).
About Trading view: thanks for your clarification.

yura505:
@Shawn8901 Yes, we a looking high/low for K last periods. Also I think you need (K-1) on line 7 if you added unclosed period there. Try (14,14,3), I hope it it what you need.

Shawn8901:
@yura505 Okay. 
Then i have one question: Isn't it the same error for rsi then?
rsi should look over the avg of last X periods, but ignores the "current" as well.

yura505:
@Shawn8901 It’s related to all indicators, they calculated for previous periods, but not current. When OnPeriod expecuted in engine.js, s.period points to current closed period and s.lookback to array of previous periods. However indicators calculates for previous periods only and no current, I think it should be fixed for all indicators.

Shawn8901:
I will invest some more research on this point. Due this i will close this pr.

PR #1055, opened 2018-01-06
Console Output adjustment

Shawn8901:
Just two adjustments for the console output which annoyed myself.

* adjusted the length of minuses in the starting header to match each other
* moved the header lines to match the report output (until custom output of the strategy)


DeviaVir (approved):
:+1: heh thanks for contributing

PR #1060, opened 2018-01-06
darwin.js - crossover strategy

shr00mie:
noticed that generational output was static from gen to gen for smalen1 and smalen2. sure enough, those are coded into the strategy itself, but were left out of the darwin.js strategy config for mutation.


DeviaVir (approved):

tiagosiebler:
@shr00mie @DeviaVir I left these out on purpose because the crossover strategy doesn't currently use these for decision making:
https://github.com/DeviaVir/zenbot/blob/unstable/extensions/strategies/crossover_vwap/strategy.js#L70

Only ema1 and vwap are currently used. So having those in the darwin parameters may only add more required tests before you start seeing improvements. The only affect they might have, is that they require more periods before the strategy kicks in, since the signal checks don't happen until all parameters are populated:
https://github.com/DeviaVir/zenbot/blob/unstable/extensions/strategies/crossover_vwap/strategy.js#L68

Realistically the strategy needs to be enhanced to either make these extra parameters an optional alternative to the ema comparison, or removing them completely.

PR #1061, opened 2018-01-06
Docker

DeviaVir:
* Resolve issues with npm install due to NPM registry issues
* Add webpack dependencies to docker before attempting `webpack -p` in `npm install`


montag:
FYI - pinkie 2.0.5 (in the lockfile) doesn't exist anymore. It appears to be from an unofficial package uploaded during a brief npm registry downtime today when the original npm registry owner (floatdrop) pulled their packages.
 
see: https://github.com/floatdrop/pinkie/issues/18

PR #1063, opened 2018-01-06
Travis. Docker.

DeviaVir:


.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: node_js
+node_js:
+  - "8"
+  - "9"
+before_install:
+  - rm -rf package-lock.json
+  - npm install https://github.com/floatdrop/pinkie-promise.git
```
DeviaVir, line 7: probably temporary, but just so we can use this now.

PR #1076, opened 2018-01-07
Post install script

kernelsndrs:


DeviaVir (approved):
:+1:

krystophv:
This won't run on Windows, which is sorta harshing the beauty of `node` being cross-platform

DeviaVir:
@krystophv talib works on Windows nowadays? I think I've given up Windows support months ago.

krystophv:
`talib` has _always_ worked on Windows. The original `node-talib` (https://github.com/oransel/node-talib) didn't have the stuff to compile it properly before making a native module via node-gyp, but there are at least two forks that do: https://github.com/drorgl/node-talib seems like the most up to date one, but https://github.com/krystophv/node-talib also works ;)

Actually, the `forex.analytics` dependency _uses_ a compiled version of `talib` that works fine on Windows, which is what led me to the fix for one of the forks.

DeviaVir:
@krystophv that's interesting, thanks for explaining that. If you find a way to make this cross-compatible I'll merge it in.

PR #1078, opened 2018-01-07
genetic_algo in python back to work

Heerpa:
by deleting one of the parameters 'period' and 'periodLength' from the list, as well as making sure different instances of Individual don't look the same (so multiple of them can coexist in a set), the algorithm could be brought back to work.


shr00mie:
any particular reason we're duplicating efforts with multiple backtesting components? would it not be wiser to pick one backtesting method and focus on optimizing that? are there features/functions which are available in one but not the other? is it a question of cross platform compatibility? seems like we've been working on/contributing mostly to darwin. if i'm missing something, please drop the knowledge.

DeviaVir (approved):
:+1: all for having actually working scripts

PR #1082, opened 2018-01-08
Update Dockerfile and docker-compose files

duanehutchins:
* Consolidate Dockerfile commands for improved caching and performance
* Change Mongodb to use a named Docker volume
* Update docker-compose to start server without calling trade --paper
* Update docker-compose-windows to match UNIX they are now the same

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,8 +2,8 @@ version: '3.1'
 
 services:
   server:
-    # image: deviavir/zenbot:unstable
-    build: .
+    image: deviavir/zenbot:unstable
+    # build: .
     volumes:
       - '.:/app'
     links:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,30 +1,34 @@
 version: '3.1'
 
 services:
-
   server:
-    image: deviavir/zenbot:unstable
+    # image: deviavir/zenbot:unstable
+    build: .
     volumes:
-      - ./conf.js:/app/conf.js
-      - ./extensions:/app/extensions
-      - ./simulations:/app/simulations
+      - '.:/app'
     links:
       - mongodb
-    command: [ "trade", "--paper" ]
+    working_dir: /app
+    entrypoint: sh -c
+    command: ["./startup.sh"]
     restart: always
     tty: true
     depends_on:
       - mongodb
     environment:
       - MONGODB_PORT_27017_TCP_ADDR=mongodb
+      - HOME=/app
 
   mongodb:
     image: mongo:latest
     volumes:
-      - ./data/db:/data/db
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -1,31 +1,35 @@
...
     image: mongo:latest
     volumes:
-      - ./data/db:/data/db
+      - data:/data/db
     command: mongod --smallfiles --bind_ip=0.0.0.0 --logpath=/dev/null
     ports:
       - 27017:27017
 
+volumes:
+  data: {}
```
DeviaVir, line 31: I'd  actually argue against this, as I'd prefer to have the data local in the same directory. Thoughts?

docker-compose.yml
```diff
@@ -1,31 +1,35 @@
...
-      - ./conf.js:/app/conf.js
-      - ./extensions:/app/extensions
-      - ./simulations:/app/simulations
+      - '.:/app'
     links:
       - mongodb
-    command: [ "trade", "--paper" ]
+    working_dir: /app
+    entrypoint: sh -c
+    command: ["./startup.sh"]
```
DeviaVir, line 13: Why not `trade --paper`?

docker-compose.yml
```diff
@@ -1,31 +1,35 @@
...
 services:
-
   server:
     image: deviavir/zenbot:unstable
+    # build: .
     volumes:
-      - ./conf.js:/app/conf.js
-      - ./extensions:/app/extensions
-      - ./simulations:/app/simulations
+      - '.:/app'
```
DeviaVir, line 8: I don't think this is a good idea, we're installing `node_modules` in the Dockerfile and this looks like it overwrites that

DeviaVir (changes requested):

docker-compose.yml
duanehutchins: Ah, yes, you're right. I'll come up with a different solution.

docker-compose.yml
duanehutchins: If trade --paper fails, the container won't start. Also, sometimes I don't want it running the trade, so this gives me the ability to trigger it myself or not.

docker-compose.yml
duanehutchins: The benefit of a named volume gives me the ability to access the volume from another heroku container. It also makes it a bit more protected from accidental deletions when a user deletes his directory.

DeviaVir: I understand the heroku use case, but I can't really get behind the accidental deletion: if I delete the zenbot dir I would really also want to delete all DB data, and not have that linger somewhere invisible to me.

PR #1088, opened 2018-01-09
adjustments to work on windows

krystophv:
switches out the `node-talib` dependency for one that's been updated to correctly build on windows and replaces a shell script with some shelljs (already a dependency of the project)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39,6 +39,11 @@
         }
       }
     },
+    "@types/node": {
+      "version": "8.5.8",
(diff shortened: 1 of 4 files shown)


krystophv:
FYI, I have a PR out against the main `node-talib` project which adds windows support without adding a number of other nested dependencies (which this currently does). Might want to wait to see if that gets merged and then I can update this to just use the new version of `node-talib`.

DeviaVir (commented):
LGTM!

Issue #1089, opened 2018-01-09
ZenbotGot stuck at sell price and constantly selling at same price at loss


code4passion:
### System information
- **Have I written custom code (as opposed to using zenbot vanilla)**: No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Linux
- **Zenbot version**: commit f235ba1e3966801b3617ac20c2ac60dcd299e8f6 (HEAD -> master, tag: v4.1.0, origin/master, origin/HEAD)
Author: Chase <chase@sillevis.net>
Date:   Sun Jan 7 02:32:12 2018 +0100
 v4.1.0: release to master (#1069)

Strategy user : trend_ema
MODE: Paper

### Describe the problem
Zenbot was trading fine in paper mode for some initially. But  after some time it started selling at exactly price at loss . This doesnot happen always, but it happend once.

<img width="1123" alt="zenbotbug" src="https://user-images.githubusercontent.com/9897265/34705758-ada5f436-f4b7-11e7-84b0-accd54896e9d.png">

code4passion:
<img width="1218" alt="screen shot 2018-01-08 at 9 09 10 pm" src="https://user-images.githubusercontent.com/9897265/34705859-57e7c3c0-f4b8-11e7-8140-7d6784f55bd4.png">

srgkmv:
This happens to me every day. I have to restart zenbot every 6 hours to avoid this bug.

SlackerATX:
I think I found this as well but only after an update in the past few days.  

How long have you been seeing this?

SlackerATX:
This seems like a new bug in the past couple of days.  I don't remember the sell price suck so off from the actual price.


> 2018-01-09 09:08:32  **14654.37** BTC-USD   +0.71%      576    -- 42    -0.1251   0.1108   +2.01%  0.00677 BTC  10132.01 USD   -3.18%   -0.57%
> manual limit SELL command executed
> 
> sell order completed at 2018-01-09 09:08:36:
> 
> 0.00670523 BTC at 14289.91 USD
> total 95.8171 USD
> 0.0000% slippage (orig. price **14289.91** USD)
> execution: a few seconds
> 
> 2018-01-09 09:13:42  14612.39 BTC-USD   +0.42%      622    -- 41    -0.1566   0.1107     sold  0.00006 BTC  10227.83 USD   -3.21%   -0.31%
> manual limit BUY command executed
[...]

PR #1098, opened 2018-01-10
Rebuild Websocket Client instead a "re"-connect

Shawn8901:
Hi,
I had the issue, that zenbot got random locked sell or buy prices. the error appeared for me after 3-4 hours every day.
As i did not find any strange behavior in the zenbot code itself i started some investigation on my exchanges code (gdax).

I found a PR in the Gdax-API repository, which enables the heart beat channel by default. For our current used version the heart beat channel is optional.

Also i somehow does not trust the "reconnect" function. For me the locked price all time appeared right after a reconnect. so i changed the onClose handler, that it builds a complete fresh websocket client instead.

Also see #1089 

After this changes no locked price issue for 23h now.

### DIFF
--- a/extensions/exchanges/gdax/exchange.js
+++ b/extensions/exchanges/gdax/exchange.js
@@ -20,7 +20,7 @@ module.exports = function container (get, set, clear) {
       try {
         auth = authedClient()
       } catch(e){}
-      websocket_client[product_id] = new Gdax.OrderbookSync([product_id], c.gdax.apiURI, c.gdax.websocketURI, auth)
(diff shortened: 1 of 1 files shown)


DeviaVir (approved):
Thanks for your debugging work, let's see it in the wild and check if it solves the issue for other as well.

blevok:
This seems to have resolved the issue for me. Thank you.

blevok:
I spoke too soon. It ran longer before it happened this time, but it did happen again. I posted details in the issue.

PR #1102, opened 2018-01-11
Add bollinger strategy to genetic backtesting

station384:
Updates the genetic backtested for the recently added bollinger strategy

### DIFF
--- a/scripts/genetic_backtester/darwin.js
+++ b/scripts/genetic_backtester/darwin.js
@@ -41,6 +41,7 @@ let iterationCount = 0;
 let runCommand = (taskStrategyName, phenotype, cb) => {
   let commonArgs = `--strategy=${taskStrategyName} --period_length=${phenotype.period_length} --min_periods=${phenotype.min_periods}  --markdown_buy_pct=${phenotype.markdown_buy_pct} --markup_sell_pct=${phenotype.markup_sell_pct} --order_type=${phenotype.order_type} --sell_stop_pct=${phenotype.sell_stop_pct} --buy_stop_pct=${phenotype.buy_stop_pct} --profit_stop_enable_pct=${phenotype.profit_stop_enable_pct} --profit_stop_pct=${phenotype.profit_stop_pct}`;
   let strategyArgs = {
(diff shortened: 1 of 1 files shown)


scripts/genetic_backtester/darwin.js
```diff
@@ -240,6 +241,23 @@ let RangeNeuralActivation = () => {
 };
 
 let strategies = {
+  bollinger: {
+    period_length: RangePeriod(1, 60, 'm'),
+    min_periods: Range(1, 20),
```
dauv, line 246: Any specific reason why you chose min_periods between 1-20? All the other strategies use 1-120, 1-200

scripts/genetic_backtester/darwin.js
```diff
@@ -240,6 +241,23 @@ let RangeNeuralActivation = () => {
...
+    markdown_buy_pct: RangeFloat(-1, 5),
+    markup_sell_pct: RangeFloat(-1, 5),
+    order_type: RangeMakerTaker(),
+    sell_stop_pct: Range0(1, 50),
+    buy_stop_pct: Range0(1, 50),
+    profit_stop_enable_pct: Range0(1, 20),
+    profit_stop_pct: Range(1,20),
+
+  // -- strategy
+    bollinger_size: Range(1, 15),
```
dauv, line 256: bollinger_size defaults to 20 in the conf.js. So I'd assume it should be included in the range aswell?

station384: thank you,   I missed that in my cleanup before check-in.   I originally had it so low as I knew anything higher in my sample data would result in 0 trades.   but you are correct,  it should be +- atleast 100 % to allow the algorithm to evolve it properly.

station384:
Removed to correct defaults for general use.

scripts/genetic_backtester/darwin.js
dauv: No problem. I had actually implemented bollinger to darwin locally, but wasn't sure on what would be the right parameter ranges.

PR #1103, opened 2018-01-11
Add bollinger strategy to genetic backtesting

station384:
Add bollinger strategy to genetic backtesting

### DIFF
--- a/scripts/genetic_backtester/darwin.js
+++ b/scripts/genetic_backtester/darwin.js
@@ -253,7 +253,7 @@ let strategies = {
     profit_stop_pct: Range(1,20),
 
   // -- strategy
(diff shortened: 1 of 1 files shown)


DeviaVir (approved):

scripts/genetic_backtester/darwin.js
```diff
@@ -240,6 +241,23 @@ let RangeNeuralActivation = () => {
 };
 
 let strategies = {
+  bollinger: {
+    period_length: RangePeriod(1, 60, 'm'),
+    min_periods: Range(1, 20),
```
dauv, line 246: @station384 What about the min_periods? It's default value is 52. Shouldn't it range between 1-120 or 1-200 like the other strategies?

PR #1106, opened 2018-01-11
add opt-in flag to use currency specific increment.

kernelsndrs:
if flag is set, utilize in order adjustment calculation.

This fixes the following scenario ( on poloniex ): 
`order status: open
2.04100007 vs our 2.04100001
`
if --currency_increment is set to .000001 ( which is the USDT increment for bids on XRP/USDT ) then the calculation wouldn't cancel and reorder.

While it would be a larger undertaking to put this into each exchanges products.json I am not sure if this increment is distributed by various exchange APIs


DeviaVir (approved):
Looks good to me, but I'll leave some time for others to comment/think about this.

PR #1126, opened 2018-01-13
 update products.json file to newest trading pairs at binance for RLC

Haehnchen:


firepol:
@Haehnchen It would be cool to have a tool (or a command, e.g. ./zenbot.sh update binance) to update the products.json automatically instead of requiring manual commits like this. The binance API already downloads all products with the call (python api): products = api.get_products()
Te job would be to parse that file and check for each pair if it already exists, if not add it to the products.json file.

Shawn8901:
@firepol Correct me if i am wrong, but the product.json is generated by the shellscript which is placed in the same folder (update-products.sh).
If you run it by yourself on a bash its all you are requesting here?

Haehnchen:
There are already scripts extensions/exchanges/*/update-products.sh, but as long the files commit here, they should be in sync ;).

So an official and documented command like `./zenbot.sh update binance` would be nice, indeed

defkev:
> ./zenbot.sh update binance would be nice

The user would still need to run it manually though.

Why not simply call the exchange specific update-products.sh before getProducts() in engine.js using exec()?

Maybe with a last-modified check of the product.json against the system time for rate limiting (e.g. only update once per hour)

firepol:
Ah I didn't know that there was a script already, actually then it would be a nice to have (just an idea, not sure it's the best for all cases) the product files to the .gitignore and automatically run the script when zenbot runs, that way the products would be always up-to-date, automatically, no need of maintenance (commits like this one) anymore, as @defkev suggests.

But it's also true that listing a new crypto in an exchange doesn't happen so often (despite a big exchange which wants to list all of 'em, like bittrex, or also binance), so in the end it's also ok to leave things as is. Like this anyway the first person who needs a specific coin and see that it's missing, runs the script and submits a pull request. Easy.

DeviaVir (approved):

PR #1127, opened 2018-01-13
Web UI fix CoinMarketCap static css 404

defkev:
Just tried the web UI for the first time and noticed that all icons were missing.

Looks like they moved some files on their end.


defkev:
Just noticed you changed the branches...

Let me try again *duh*

PR #1128, opened 2018-01-13
Allow the REST API/Web UI to run on a specific IP

defkev:
Instead of listening on all available interfaces

e.g. use:
c.output.api.ip = '127.0.0.1' in conf.js
to bind to localhost (for use with a reverse proxy)


conf-sample.js
```diff
@@ -241,4 +241,5 @@ c.output  = {}
 // REST API
 c.output.api = {}
 c.output.api.on = true
+c.output.api.ip = // either a IPv4 or IPv6 address (as a string) to listen on, uses all available interfaces if omitted
```
kernelsndrs: this should probably default to 0.0.0.0 or 127.0.0.1 as if the setting isn't set in the conf.js it falls back to conf-sample.js and currently I believe this would leave it undefined

DeviaVir: shouldn't this be an empty string, or null?

defkev: I want to avoid breaking existing installations thus its undefined as api.js takes care of it anyway.

kernelsndrs: I commented the same thing but realized api.js handled it as well.  I think that if it was set to "0.0.0.0" it wouldn't break existing installations as well as function as an opt-in parameter that the user has to change in their own conf.js

defkev: Sounds like the most elegant solution as it also makes it clear how the settings is supposed to look like.

DeviaVir (approved):

PR #1132, opened 2018-01-14
Fix output location for the HTML result dump

firepol:
Trade results were saved in the simulations/ folder. To avoid confusion,
paper results are prefixed with "paper" (& still saved in simulations/).

### DIFF
--- a/commands/trade.js
+++ b/commands/trade.js
@@ -217,16 +217,16 @@ module.exports = function container (get, set, clear) {
               .replace(/\{\{symbol\}\}/g,  so.selector.normalized + ' - zenbot ' + require('../package.json').version)
             if (so.filename !== 'none') {
               var out_target
-              
+              var out_target_prefix = so.paper ? 'simulations/paper_result_' : 'stats/trade_result_'
               if(dump){
                 var dt = new Date().toISOString();
                 
                 //ymd
                 var today = dt.slice(2, 4) + dt.slice(5, 7) + dt.slice(8, 10);
-                out_target = so.filename || 'simulations/trade_result_' + so.selector.normalized +'_' + today + '_UTC.html'
+                out_target = so.filename || out_target_prefix + so.selector.normalized +'_' + today + '_UTC.html'
               fs.writeFileSync(out_target, out)
               }else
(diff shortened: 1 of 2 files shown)


commands/trade.js
```diff
@@ -217,16 +217,16 @@ module.exports = function container (get, set, clear) {
               .replace(/\{\{symbol\}\}/g,  so.selector.normalized + ' - zenbot ' + require('../package.json').version)
             if (so.filename !== 'none') {
               var out_target
-              
+              var out_target_prefix = so.paper ? 'simulations/paper_result_' : 'stats/trade_result_'
```
DeviaVir, line 220: `stats/trade_result`?

firepol: `trade_result_` was already the prefix of the file name. Stats, because I was checking the stats folder: https://github.com/DeviaVir/zenbot/tree/unstable/stats

The readme says *HTML Outputs from trades will be placed here!*
I've run several live modes and they were saved in the simulations folder, which doesn't make sense. They should be saved in the stats folder (even he readme says so, or did I misunderstood something?).

I added the variable `out_target_prefix` which is used a few lines below.

Actually if you are ok with this change, I forgot to add paper_result* in the .gitignore:

Before:

```
sim_result*
trade_result*
```

After:

```
sim_result*
trade_result*
paper_result*
```

I'm still new to github pull requests. For such a thing what should I do? Close this pull request and create a new one that includes also the .gitignore update so we have just one commit with everything together?

dauv: @firepol Regarding your pullrequest question. If you want everything in a single commit. You can do the following:

First reset the HEAD (add a ^ for each commit you want to reset)  of your local branch:
```
git reset HEAD^
```

Then make your change and commit again. And force push to your remote branch, to overwrite the commtis in the pullrequest:

```
git push stats --force
```

firepol: Thanks @dauv, I added `paper_result*` to the gitignore, reset my commit and forced pushed to it, github automatically updated it. Cool stuff.

DeviaVir (approved):

PR #1143, opened 2018-01-15
Improve truthy & falsey handling of options which could be set to 0

### LINKED ISSUE #1141: Bug: zenbot sells in loss even if configured otherwise
There is an option called `max_sell_loss_pct`.

In the `conf-sample.js` I read:

```
// avoid selling at a loss below this pct set to 0 to ensure selling at a higher price...
c.max_sell_loss_pct = 25
```

so I configured as follows:

`c.max_sell_loss_pct = 0`

I did set it to 0. I did run a zenbot session in live and still zenbot bought e.g. at 0.00015 and sold some hours later at 0.00012. This is quite a big percent of loss. For a volatile crypto it's ok for the price to go up and down, what I definitively don't want is the bot to sell if I configured `c.max_sell_loss_pct = 0`.

This looks like a critical bug to me.

To reproduce I guess you can backfill some data and test with the sim. You should not get any losses if configured to 0, else the option is misleading and should be documented otherwise.

defkev:
Fixes #1141

### DIFF
--- a/lib/engine.js
+++ b/lib/engine.js
@@ -460,9 +460,9 @@ module.exports = function container (get, set, clear) {
               if (s.product.max_size && Number(size) > Number(s.product.max_size)) {
                 size = s.product.max_size
               }
-              if (s.buy_order && so.max_slippage_pct) {
+              if (s.buy_order && so.max_slippage_pct != null) {
                 var slippage = n(price).subtract(s.buy_order.orig_price).divide(s.buy_order.orig_price).multiply(100).value()
-                if (so.max_slippage_pct && slippage > so.max_slippage_pct) {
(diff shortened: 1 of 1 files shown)


lib/engine.js
```diff
@@ -460,9 +460,9 @@ module.exports = function container (get, set, clear) {
               if (s.product.max_size && Number(size) > Number(s.product.max_size)) {
                 size = s.product.max_size
               }
-              if (s.buy_order && so.max_slippage_pct) {
+              if (s.buy_order && so.max_slippage_pct != null) {
```
DeviaVir, line 463: shouldn't this be `!== null`?

defkev: Short version of `!== null || !== undefined`

DeviaVir (approved):

PR #1147, opened 2018-01-15
Update bitstamp products (add BCH & fix min_sizes)

firepol:
Add bitcoin cash & fix the minimums (5 USD/EUR or 0.001 BTC)
See https://www.bitstamp.net/api/v2/trading-pairs-info/

### DIFF
--- a/extensions/exchanges/bitstamp/products.json
+++ b/extensions/exchanges/bitstamp/products.json
@@ -3,7 +3,7 @@
     "id": "BTCUSD",
     "asset": "BTC",
     "currency": "USD",
-    "min_size": "0.01",
+    "min_size": "5",
     "max_size": "10000",
     "increment": "0.01",
     "label": "BTC/USD"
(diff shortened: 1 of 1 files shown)


DeviaVir:
This PR will be overwritten (in case it was done manually) by our travis cronjob. 
https://github.com/DeviaVir/zenbot/blob/unstable/scripts/cron.sh

firepol:
That's ok, I just didnd't see a script to update bitstamp products so I was thinking bitstamp was the only manual one for now.

Btw I hope that the script will update the min_size amounts correctly. The old products.json had wrong values, the minimum trade on bitstamp as I wrote in my commit message is 5 USD, 5 EUR or 0.001 BTC.

Maximum amounts I have no idea, but I guess the number used is large enough. Eventually some whales one day will complain ;)

DeviaVir:
Ha, whoops, you're right. I didn't put 1 and 1 together (the enhancement ticket for bitstamp and this PR). Will merge it.

DeviaVir (approved):

PR #1167, opened 2018-01-18
Momentum: fix strat name

DeviaVir:
Resolve problem reported in https://github.com/DeviaVir/zenbot/pull/1163

This worked with command (just an example):
```
./zenbot.sh sim gdax.BTC-EUR --strategy momentum --days=5 --period=10m --asset_capital=10 --currency_capital=0 --avg_slippage_pct=1
```


hig-big:
worked. thanks.

PR #1168, opened 2018-01-18
Fix for RSI returning NaN values

aquamanjl:
Under certain conditions the current RSI lib returns NaN values. Added check for divide by zero condition and put in test cases to validate changes.


DeviaVir (approved):
:+1: also definitely appreciate the added test case!

PR #1171, opened 2018-01-18
Web UI: Replace tradingview.com widget with trade chart using our own data

defkev:
Uses plotly.js (bundled) and the REST API to plot price, volume, OCHL (using lookback) as well as buy/sell signals.

Not super pretty but gets the job done:

plotly


DeviaVir (approved):
:+1: looks good to me, I'd rather have us use our own tools than include tradingview.

lib/engine.js
```diff
@@ -149,6 +150,7 @@ module.exports = function container (get, set, clear) {
       s.period.close_time = trade.time
       s.strategy.calculate(s)
       s.vol_since_last_blink += trade.size
+      s.trades.push(trade)
```
kernelsndrs, line 153: Don't we think this might start bloating the s object?  In the same vein as the s.lookback bloating and what I assume was the cause for any Out Of Memory errors.

kernelsndrs (commented):
Not sure if this is going to cause bloating issues.

evseevnn-zz:
im so late...
Why not use [echarts](https://ecomfe.github.io/echarts-examples/public/index.html) ?
They really better...

lib/engine.js
defkev: > cause for any Out Of Memory errors

Did i miss something? Never seen any of my bots go OOM yet.

Some (really) rough size calculation on s.trades:
`[trades] 505, [size] 86366 bytes, [size/trade] 171 bytes`

Assuming an average size of 171 bytes per trade would mean 10000 trades "only" taking up 1.63 Mb which imho is negatable.

defkev:
@evseevnn
Looks pretty neat

My biggest concern with plotly.js right now is that performance drops drastically with many datapoints.
How does echarts hold up in this regard?

lib/engine.js
kernelsndrs: https://github.com/DeviaVir/zenbot/issues?q=is%3Aissue+out+of+memory+is%3Aopen

So what if you aim to have this bot running for a month or more? Do you still think its going to be 10000 trades? While I'm sure that it could be considered negligible in the short term, I think it could also be considered sloppy to not just add a limit and unshift them..

defkev: Depends on the pair/exchange you are trading/trading on.

All the issues seem to be somewhat related to either running in docker or mongodb...i wonder if zenbot has a general problem in low memory environments?
Let me spin up a test box and see what happens.

evseevnn-zz:
@defkev Sorry but i dont know.
I long time wanna try this library but have no time for that.

lib/engine.js
kernelsndrs: https://github.com/DeviaVir/zenbot/issues/1185

Seems interesting / coincidental that an issue such as this presents itself..

defkev: Its all just a coincident 😜 

I deployed a nano instance yesterday (1 vCPU/512MB RAM) and the bot is still running rock solid at a steady mem util of 10% - 12%, only the webpack bundling took almost 10 minutes.

We could stop s.trades from populating if `c.output.api.on = false` (as it is only needed by the API anyway) for people trying to run zen on a potato.

defkev: See https://github.com/defkev/zenbot/commit/7102a512a413cf34d29cbd2f6a0692ec2c99cc29

kernelsndrs: I think that's a great idea to help accommodate those on a potato 👍

PR #1194, opened 2018-01-21
Fix #1184 cexio update-products.sh

### LINKED ISSUE #1184: cex.io wrong products.json and issue with update-products.sh
How to reproduce (under linux), latest unstable branch:

```
cd extensions/exchanges/cexio
./update-products.sh
```

What I get:
`/usr/bin/env: ‘node\r’: No such file or directory`

@DeviaVir can you please fix the script and included it in the nightly, so the products.json will be updated automatically?

Some issues I've found (and wanted to fix):
- min_size often wrong
- some labels wrong, e.g. XRP to EUR (and GBP) label is still XRP/USD

firepol:
This fixes the error `/usr/bin/env: ‘node\r’: No such file or directory`

### DIFF
--- a/extensions/exchanges/cexio/update-products.sh
+++ b/extensions/exchanges/cexio/update-products.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env node
+
 const ccxt = require('ccxt')
 const path = require('path')
 var client = new ccxt.cex()


DeviaVir (approved):
:+1:

PR #1196, opened 2018-01-22
Fixed StochRSI calculations

aquamanjl:
Main changes in this PR include

- Fix the calculations for Stochstic RSI module

- Convert RSI and SRSI output to floats to better match expected values

- Test case for SRSI


DeviaVir (approved):
:+1: thank you for your contribution

Issue #1195, opened 2018-01-22
Bollinger High/Low Status


station384:
### System information
- **Have I written custom code (as opposed to using zenbot vanilla)**: Yes
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Linux Ubuntu 16.04 / macOS Current Version
- **Zenbot version**: unstable
- **NodeJS version**: current ubuntu release
- **Python version**: 
- **Exact command to reproduce**: ./zenbot sim gdax.bch-btc --strategy bollinger --days=3 

### Describe the problem
Bollinger High / Low status display only show whole numbers,  This results in confusing values when trading on values with a price less than 1.  i.e.  bch-btc,  1 bch is currently trading at 0.1XXXX this results in a High/Low value of 00 00  where it should display 00.1XXX 00.1XXX.
Looking thru the code there is room for 7 characters and one space for each high and low value so should not cause any issues. 

This effects Bollinger and Trend_Bollinger

station384:
I'll make this change myself, if feedback is good.

DeviaVir:
Sure, go for it @station384 :)

PR #1198, opened 2018-01-22
Bollinger High/Low enhancement

station384:
Report is now able to display values less than 1.  space limited to 7
characters per column so smallest value is 0.00001
enhancement #1195


DeviaVir (approved):
:+1:

Issue #1199, opened 2018-01-22
Sim command save config to file


station384:
TL;DR
 Add parameter to sim command.  outputs json file with sim results. Modify backtesters to use said json file.

The change comes from working with the genetic_backtester and some errors that occur when it scrapes the JSON results from stdio.   
Occasionally the regex will fail with index 1 not found.
This error only seems to occur after a period of time,  it is rather random. When debugging, it seems the failure is in the regex when decodes the first match.   reworking/debugged the regex,  filtering the text before the regex doesn't seem to correct the problem, I have a theory that it is some control character causing it to fail. 
 
I propose and alternative solution; writing out the Config JSON output to a file after the sim completes with an additional item added to the json called simresults which has all the results (end currency, trades, etc) from the sim.

I have a modification to the sim command that adds a new parameter called **backtester_generation**. It accepts a number.  if the number is greater than 0 it writes the json file in the format of <selectedexchange.marketpair>-<backtester_generation>.json

This file can then be read by the backtesters instead of scraping using regex, this will provide more reliability and eliminate the possibility of extra text in the stdio causing problems. 

I also see an alternative approach where we store the results in mongodb, which I think should be the final direction, which would allow for easier continuation of generations instead of the current setup using json files, the downside is a lot of extra data bloating the database.

Looking for comments,  thoughts,  other ideas of what could be done with this, or nah not needed.

Issue #1131, opened 2018-01-14
Save history of executed trades (trade & sim mode) in a json file


firepol:
Now, the simulations are saved in the simulations folder as html file called e.g. sim_result_exchange.XYZ-BTC_180113_195655_UTC.html

trades are also saved in the simulations folder, but the file has a name e.g. trade_result_binance.XYZ-BTC_180114_UTC.html which I think should be saved under "stats" (I think this is a minor bug, I think I'll fix that: paper trades should be saved under simulations, but real trades under stats).

When opening the HTML I see many little green arrows and red arrows showing when zenbot did buy and when it sold.

What I'd like to see is to see a history of executed trades, at best as a JSON file.
In paper mode I can receive a notification on telegram to know what the bot would have done, so it's possible to append the data also to a json.

Ideally it should show everything, also failed attempts to buy e.g. when placing a limit order that was not filled. So e.g. the object should have a property called "executed" : true, false if not executed (so you see probably another request a few seconds later where executed is true, because thew previous one didn't go through, is this how the slippage works?)

The HTML should load the json and display the history at the bottom, just after "error rates".

**The advantage of this**:
- better understanding on how zenbot works
- spot issues / bugs
- easier sharing (when reporting a bug)
- **testing/unit testing**: I've seen @T-rav started adding tests, tests are a great idea and if results are saved in a json it's possible to make tests and modify the engine with confidence. If a test breaks then we know that something went wrong when fixing this or that
- tests can parse the json (and ignore the HTML)

firepol:
@defkev I see that you just submitted a pull request to load previous trades: https://github.com/DeviaVir/zenbot/pull/1243

So I was wondering if that could help maybe in regards of this suggestion?

Are previous trades saved only in live mode or also in simulation / paper?
Because my suggestion here is to save them on file, but if they are saved in the DB, then this should be an easy task?

Thought / ideas? Thx for sharing

defkev:
Trades are only saved/loaded in live/paper mode.

Including sim would clutter up the mongo collection as each simulation would generate more and more new trades, including possible duplicates, we definitely don't want to pop up in live mode.

Any specific reason you want to store this from a simulation?

I am personally more interested in the result of a sim run than its trade history 😄 

[...]

PR #1206, opened 2018-01-23
Output json result to file during simulation

station384:
See issue #1199

JSON file is implemented in genetic_backtester


DeviaVir (approved):
LGTM, will allow space for others to comment as well.

firepol:
@station384 thanks for this. Since you worked on saving information to a JSON file, MAYBE can you consider this? https://github.com/DeviaVir/zenbot/issues/1131

Or comment here or in my post some draft / hints so if you have no time at least I can try to implement my feature request? Sharing knowledge always helps.

Also, is the produced JSON in this commit in the `.gitignore`? If not please add it, else everytime you add a simulation it'll add more files you'll have to be careful not to include in future commits...

Thanks

station384:
I'll take a look at #1131. 
It may be the weekend before I can get to really checking it out.  Work calling,  I can no longer procrastinate.....

station384:
Found an error in the new vsBuyHold calculation.  correcting shortly.

station384:
issue #1210 addressed in last update.

Bug fixes only now, or I will keep tweaking it and it will never get merged :)

scripts/auto_backtester/backtester.js
```diff
@@ -16,11 +16,12 @@ let parallel  = require('run-parallel-limit');
 let json2csv  = require('json2csv');
 let roundp    = require('round-precision');
 let fs        = require('fs');
+let path = require('path')
 let StripAnsi = require('strip-ansi');
 
 let VERSION = 'Zenbot 4.04 Backtester v0.2';
 
-let PARALLEL_LIMIT = require('os').cpus().length;
+let PARALLEL_LIMIT = 3//require('os').cpus().length;
```
micheledicosmo: ^^

DeviaVir: why not make this depending on the cpus?

micheledicosmo (commented):
*comment deleted*

scripts/auto_backtester/backtester.js
micheledicosmo: If I have to take a guess this was introduced for debug purposes and then forgotten to undo

station384: correct.   it wasn't supposed to be checked in.  Still figuring out git.  totally breaks my source control flow ( used to SVN )

DeviaVir (approved):

scripts/auto_backtester/backtester.js
firepol: @station384 since you are new to git, I assume you are also a bit new to github, like me. Just sharing some info that I figured out also in one of my first pull requests... what I did is to merge changes from the upstream branch to my fork, commit one change and finally do a pull request comparing my fork with the official repo/branch. As I noted here, the pull request included lot of commits I didn't do. Somehow this is a weird behavior of github.

To avoid that (next time, now too late since it was merged already), at best you keep working on your branch, but just a few minutes before making the pull request you create a new local branch based on the upstream one (I try to call such branches with the name of what they do, or issue id... but the name doesn't matter). Then you cherry pick your changes or merge and make just one commit, and push your branch new branch containing just that one commit.

Then, when you do the pull request and choose the branch you just pushed in your repo, github will compare that branch with the actual one, where you want your changes to go. Like this, github picks only your changes and nothing else added in the meantime by other developers, and keeps the commit message clean (only your commit message).

I had the same issue, if you see your commit now it seems you did so many changes, but this is because github included lot of stuff really unrelated to your change...

@DeviaVir I think that this should be documented, if not in the readme, there should be a readme for contributors with guidelines to follow.

I see that there is a code of conduct, maybe it makes sense to create a DEVELOPMENT.md file or something similar, containing also such hints as the one I wrote above, useful also for people new to github (this issue isn't really documented in github)...

DeviaVir: @firepol usually one would add a `CONTRIBUTING.md` but I don't think you'll find basic git usage in that document, usually. Would not be a bad place for it though.

The reason we don't have a `CONTRIBUTING.md` is because it's really a lot of writing and setting up general rules, which I simply haven't had the time yet to do. When I do have a bit of spare time I prefer to review PR's and play around with strats/scripts/ideas vs. writing up documentation, if you know what I mean.

If anyone reading this feels inclined to jump in and write this document, the guidelines are available here:
https://help.github.com/articles/setting-guidelines-for-repository-contributors/

station384: I sort of figured out the branching part, but wasn't sure, for every change I want to push I was making branches off of my fork of unstable,  the largest problem I've had is when I've committed something, and published.  in SVN I would just check out from a prior check in and recheck in.  I assume GIT is similar.  

I also get messed up because in svn, I would just check in on, and then there is a separate commit where I can cherry pick the files I commit (I usually have 4 or more mods going at once in the same branch in SVN). 
My SVN process seems to be what the branching in git does.

Git has more planning ahead of what you want to commit, where as SVN happens later.

Sooner or later I'll catch on, I've been using SVN for the last decade so old habits die hard, and the stuff I learned on git from about 2 years ago has to be relearned :/  

But thanks for the tips,   confirms a couple of things, and lay off the batch commit of files like I have been doing and start doing them very specifically as you described.

and "the pull request included lot of commits I didn't do. Somehow this is a weird behavior of github"  this was bit alarming, to the point of me redoing branches and such.   
I was reading up on it and it appears you can do something with rebase to eliminate them,  essentially moving your point forward,  but I've yet to figure it out.

DeviaVir: @station384 for your last bit, if you have no changes in your fork's `unstable` branch you can do:
```
git reset --hard upstream/unstable
git push --force
```

Then any time you branch off of your fork's unstable it won't have all the changes, as the head has been moved forward.
Unfortunately, you'll need to do this every time upstream (hey, that's me) updates.

station384: TYVM  that is exactly what I was looking for. 
It has been driving me nuts.

Now to make that command into a git alias :)

PR #1207, opened 2018-01-23
Add Visual Studio Code launch settings

station384:
Adds some default settings for Visual Studio Code to allow for integrated debugging for anyone that uses it. 

Not sure if anyone else uses this ide, or if we want to put in defaults for different editors.


haxwell:
My initial thought is that IDE settings like this should not be in the source repo, FWIW.

station384:
Yeah I'm on the fence about it. 
In this case it holds only debug settings for launching tasks so its easier for devs to get started debugging, testing ext.  Only reason I thought others might like it because if they use it, it would save them the time I already spent on figuring out the settings to get it to launch.

I'm totally against IDE preference settings being in the repo.

station384:
Grrr..   Git is my nemesis.

DeviaVir:
@station384 you can totally add it in the `docs` folder though, I'd gladly accept that PR.

PR #1214, opened 2018-01-24
Refactor backfill for unit testing

haxwell:
No functionality change, other than UI, and that is minor. The code has been rewritten with an eye toward DRY, SOLID, code. Take note of the 

- Consume-And-Process Service, which provides a structure for batch processing.
- Resume-Marker Service, manages resume markers in memory, and flushes the new state with each batch.
- CollectionsService, which uses direct Mongo DB objects, rather than Carlos' SOSA package. I just didn't see the value in SOSA. Can someone enlighten me?
- The Stub Exchange, can we model all exchange objects this way? Its effectively the same as before, just cleaner. 
- Unit Tests and Mock Service Factories, how cool are mocks?

Please test test test!


haxwell:
GDAX is the only exchange that seems to offer a reliable backfill source. In my testing, the other exchanges, fail in some way or another. 

If you have access to other exchanges, please test this PR, and let me know if they are working as expected for you. Thanks!

DeviaVir (approved):
initial readthrough LGTM, I'll run some tests as well

:+1: on the tests 🥇

commands/backfill/backfill.update-screen.function.js
```diff
@@ -0,0 +1,16 @@
...
+
+module.exports = (function (get, set, clear) {
+
+  	var exchangeService = get('lib.exchange-service')(get, set, clear)
+
+	return (trade_id, data) => { 
+
+		if (data !== undefined && typeof process.stdout.clearLine == 'function') {
+			process.stdout.clearLine();
+			process.stdout.write("Processed trades up to " + Moment(data.time).fromNow() + "." );
```
DeviaVir: do you think we can show here exactly how many trades we have processed? I think that might give a better sense of what the bot is exactly doing (and what is taking so long)

haxwell: Good call.. I will make that change soon. I was going to have it display the time that the last trades were received/processed, too.

haxwell: Done in 5261f02 !

haxwell:
My earlier comment is collapsed above, but just so its out there, I added the bit about showing the number of trades processed during backfilling in 5261f02

DeviaVir:
My tests look good :+1: merging

Haehnchen:
fyi this merge breaks backfill in some way and so also live trading warmup:

new:
{ product_id: 'EOS-BTC' }
{ product_id: 'EOS-BTC', from: '4558738' }

old:
binance.EOS-BTC saved 1814 trades 1 days left
{ product_id: 'EOS-BTC', from: 1517569760767 }
.{ product_id: 'EOS-BTC', from: 1517572758190 }

there is something wring in starttime and endtime calcualtion. noticed via binance exchange, but possible an overall problem?

dakreepy1:
I am also seeing this one breaking backfill as well.  running backfill on gdax takes a very long time.  When i use the trade option and it backfills the recent trade it pretty much sits saying backfilling....
i've opened an issue with my results a few minutes ago.

Haehnchen:
because it starts backfiling somewhere around 1970 :)

haxwell:
Noted... I will look into the issues tonight.

station384:
see #1271

haxwell:
See #1279.

Issue #1211, opened 2018-01-23
Crash during live trading.


station384:
Branch: Unstable

Getting this error.  Looks like it could be that the api returned an error that was unknown.
I've received this error 3x in the last 5 hours. 

Command:
./zenbot.sh trade bittrex.XRP-USDT --avg_slippage_pct=0.045  --bollinger_lower_bound_pct=16.573725742046623 --bollinger_size=5 --bollinger_time=5.718442655882575 --bollinger_upper_bound_pct=6.55491217292086 --buy_stop_pct=12 --markdown_buy_pct=0.6928484993307022 --markup_sell_pct=-0.0014423681389312826 --max_sell_loss_pct=25 --max_slippage_pct=5 --min_periods=20 --order_type=maker --period=16m --period_length=16m --profit_stop_enable_pct=0 --profit_stop_pct=9 --rsi_periods=14 --sell_stop_pct=36 --strategy=bollinger

/extensions/exchanges/bittrex/exchange.js:162
        if(!data.success) {
                 ^

TypeError: Cannot read property 'success' of null
    at /media/zenbot/zenbot/extensions/exchanges/bittrex/exchange.js:162:18
    at Request._callback (/media/zenbot/zenbot/node_modules/node.bittrex.api/node.bittrex.api.js:127:11)
    at self.callback (/media/zenbot/zenbot/node_modules/request/request.js:186:22)
    at emitOne (events.js:116:13)
    at Request.emit (events.js:211:7)
    at ClientRequest.<anonymous> (/media/zenbot/zenbot/node_modules/request/request.js:785:18)
    at Object.onceWrapper (events.js:313:30)
    at emitNone (events.js:106:13)
    at ClientRequest.emit (events.js:208:7)
    at TLSSocket.emitTimeout (_http_client.js:708:34)
    at Object.onceWrapper (events.js:313:30)
    at emitNone (events.js:106:13)
    at TLSSocket.emit (events.js:208:7)
    at TLSSocket.Socket._onTimeout (net.js:407:8)
    at ontimeout (timers.js:475:11)
    at tryOnTimeout (timers.js:310:5)
    at Timer.listOnTimeout (timers.js:270:5)

defkev:
`node.bittrex.api` is EOL and has be superseded by `node-bittrex-api`

See https://github.com/n0mad01/node.bittrex.api and https://github.com/dparlevliet/node.bittrex.api

Updating package.json should suffice as it's just a fork.

[...]

Issue #1170, opened 2018-01-18
Problem with Bittrex exchange.js:113


sembsa:
### System information
- **macOS 10.13.2**:
- **Zenbot 4.1.0**:
- **NodeJS 9.3.0**: 
- **Python 2.7.14**:
- **zenbot trade bittrex.ETH-BTC **:

### Describe the problem
Program error condition after one hour of work.

### Source code / logs
zenbot/extensions/exchanges/bittrex/exchange.js:113
        if(!data.success) {
                 ^

TypeError: Cannot read property 'success' of null
    at /Users/sembsa/Desktop/zenbot/extensions/exchanges/bittrex/exchange.js:113:18
    at Request._callback (/Users/sembsa/Desktop/zenbot/node_modules/node.bittrex.api/node.bittrex.api.js:127:11)
    at self.callback (/Users/sembsa/Desktop/zenbot/node_modules/request/request.js:186:22)
    at Request.emit (events.js:159:13)
    at ClientRequest.<anonymous> (/Users/sembsa/Desktop/zenbot/node_modules/request/request.js:785:18)
    at Object.onceWrapper (events.js:254:19)
    at ClientRequest.emit (events.js:159:13)
    at TLSSocket.emitTimeout (_http_client.js:710:34)
    at Object.onceWrapper (events.js:254:19)
    at TLSSocket.emit (events.js:159:13)
    at TLSSocket.Socket._onTimeout (net.js:421:8)
    at ontimeout (timers.js:466:11)
    at tryOnTimeout (timers.js:304:5)
    at Timer.listOnTimeout (timers.js:264:5)

station384:
This fix for this has been merged into the unstable branch.

Issue #1164, opened 2018-01-17
Bittrex 'Error during syncing balance'


brucetus:
### System information
- **Have I written custom code (as opposed to using zenbot vanilla)**: No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Debian 9
- **Zenbot version**: Latest 
- **NodeJS version**:  8.9.4
- **Python version**: ?
- **Exact command to reproduce**: ./zenbot.sh trade bittrex.ADA-BTC 

### Describe the problem
New error popping up multiple times related to syncing balances and API keys

### Source code / logs
I'm running multiple pairs in different screens. I've been getting this error recently which I haven't gotten before, so it may be related to recent changes. It seems to have something to do with my API key, but I've checked multiple times and it is correct in the conf.js and the permissions are correct. This is related to issue #1085 , which has been closed by the creator, but doesn't give a solid fix. Also he seemed to get the error instantly after starting up the bot while mine happens hours after. I think an easy fix would be to put in an line that prevents the bot from crashing if the API can't call the account balance, either around line 506 of the trade.js file or somewhere in Bittrex's exchange.js file. If anybody can think of a way to do this let me know. Thanks.

<img width="896" alt="screen shot 2018-01-17 at 10 47 55 am" src="https://user-images.githubusercontent.com/31704711/35061505-4dfb41f6-fb76-11e7-9337-ec61247769e0.png">

station384:
Are you running all instances at 1 minute intervals? 
(Below is my findings,  I have found no official statements from bittrex)
I ask because if bittrex is polled to fast it gives odd errors, one of them I've noticed is this error.  
I don't know this for sure but from what I have noticed, there is a max polling interval for each IP address connected to the server,  so if your behind a NAT, the combined polling of all clients can result in bittrex sending out errors 500 or 504,  I've also gotten this when there site crashes.

brucetus:
Im running with a 4h period_length and my polling options are all default. Eventually every instance I was running on Bittrex crashed with the same error :(

brucetus:
Eventually every instance I was running on Bittrex crashed with the same error.

tskazinski:
I'm also getting this error with Kraken:
error syncing balance / RequestError: Request timed out
Happens every couple mins, my period is 1m though.

station384:
@brucetus, the fix for this has been merged into the unstable branch

Issue #499, opened 2017-08-19
Bittrex Backtesting data


webworker01:
Hello, I found that there is a way to get backtest data for Bittrex using this [API endpoint](https://bittrex.com/Api/v2.0/pub/market/GetTicks?marketName=BTC-ETH&tickInterval=fiveMin):

From [this reddit thread](https://www.reddit.com/r/BitcoinMarkets/comments/6k75ue/unable_to_get_historical_bittrex_data/) I found that the interval can be modified with the following params:
["oneMin", "fiveMin", "thirtyMin", "hour", "day"]

Hopefully this can be added in at some point if someone finds the motivation and time 👍

webworker01:
I was poking around and see that [this client](https://github.com/n0mad01/node.bittrex.api/blob/master/node.bittrex.api.js#L278) has already implemented the method.

station384:
Interesting.... I'll look in to this and see if it can be done since I've been poking around in bittrex lately.

PR #1215, opened 2018-01-24
Bittrex Exchange

station384:
Correction for #1211, #1170, #1164 
Updated bittrex api from node.bittrex.api -> node-bittrex-api

node.bittrex.api is no longer maintained

Added some error handling so exchange is now passing back useful errors
to the engine.
i.e.  it will no longer keep retrying a buy sell order if there isn’t
enough funds.

socket errors no longer crash the system but passed to the retry
handler.

**npm must be updated.
npm install node-bittrex-api**

After these corrections it eliminated the errors cited in #1211. 

Unintentional corrections: (I like these :)
Errors now passed back to engine.js for proper handling
Orders will now readjust price ( old versions was silently failing )


extensions/exchanges/bittrex/exchange.js
```diff
@@ -1,9 +1,8 @@
-var bittrex_authed = require('node.bittrex.api'),
-    bittrex_public = require('node.bittrex.api'),
-    path = require('path'),
-    moment = require('moment'),
-    n = require('numbro'),
-    colors = require('colors')
+var bittrex_authed = require('node-bittrex-api'),
+  bittrex_public = require('node-bittrex-api'),
+  moment = require('moment'),
+  n = require('numbro')
```
DeviaVir: nit: indenting?

extensions/exchanges/bittrex/exchange.js
```diff
@@ -1,9 +1,8 @@
-var bittrex_authed = require('node.bittrex.api'),
-    bittrex_public = require('node.bittrex.api'),
-    path = require('path'),
-    moment = require('moment'),
-    n = require('numbro'),
-    colors = require('colors')
+var bittrex_authed = require('node-bittrex-api'),
```
DeviaVir: Should `historyScan` still be `forward`?
https://github.com/DeviaVir/zenbot/pull/1215/files#diff-b3b56e634e6ef011a9923a3c54c40108L51

extensions/exchanges/bittrex/exchange.js
```diff
@@ -200,26 +233,50 @@ module.exports = function container(get, set, clear) {
...
-          return cb(null, [])
+          return cb(null, {})
         }
-
+      
+      
+ 
         if(!data.success) {
-          if (data.message && data.message.match(recoverableErrors)) {
+          if ( data.message && data.message.match(recoverableErrors)) {
```
DeviaVir: nit: unnecessary whitespace

DeviaVir (approved):

station384:
will correct in a few.

extensions/exchanges/bittrex/exchange.js
station384: Unsure.   The market history  on bittrex is very limited.   It will pull only the last couple of trades.  there is a workaround that someone came up with using the v2 api that seems to get around this limitation,  I'll look at implementing it.  hate having to leave a marking noop trading for days to get a decent history on.

extensions/exchanges/bittrex/exchange.js
station384: It is formatted according the the setting in .eslintrc.
I'll manually format it.

extensions/exchanges/bittrex/exchange.js
station384: I just double checked this   moving to forward causes 
on markets I already have data on
error: getTrades() returned duplicate results
{ product_id: 'XRP-USDT', to: 1516785694047 }
{ product_id: 'XRP-USDT', to: 1516785694047 }
it returns nothing on markets I don't have data on.

So for now, I think it should remain as backward until I can looking into the history calls in bittrex.

station384:
@DeviaVir, it looks like I got all the corrections for formatting.   

The history part is just broken due to the api, and the way bittrex returns market data. 
I'll look at that later today and see if I can implement the a possible fix mentioned in #499.

I would also like to convert this to using websockets but that is for a later date.

DeviaVir:
SGTM I will merge this now so we can look at the other additions fresh.

PR #1227, opened 2018-01-26
CEXIO update products fix

nedievas:
to be working after January 23, 2018 official exchange trade engine update.
More [info](https://blog.cex.io/news/engine-updates-16956)

Execute this file after `git pull`.

PR #1228, opened 2018-01-26
Update CEXIO products.json

nedievas:
#1227


DeviaVir:
Is this the result from an `update-products.sh` call?

If yes: travis should automatically pull in the changes (every 24 hours)
If no: travis will overwrite these changes

edit: context is https://github.com/DeviaVir/zenbot/pull/1227, so now this makes sense.

DeviaVir (approved):

PR #1240, opened 2018-01-28
Web UI: Replace plotly.js with echarts

defkev:
After a couple of days uptime the dashboard takes minutes to load, possibly even freezing the whole browser for a few seconds.
I have no idea why plotly.js is struggling so hard with just a few thousand datapoints and i don't even want to know what happens once we reach 10k...i suppose they have a serious memory leak somewhere. 

Functionality is pretty much the same, just another library.

After running it for over a week i can indeed confirm it performs a lot better compared to plotly.

dashboard


evseevnn-zz:
Look strange)) Can we use just japanese candle?

PR #1241, opened 2018-01-29
Replace inline javascript in simulation result with external cdn

Haehnchen:
Will reduce file size alot and allows browser caching of external resources


DeviaVir (approved):
:+1:

PR #1251, opened 2018-01-31
fix phenotype win loss ratio weighting

krystophv:
Current calculation in the phenotype fitness are incorrect. The `wlRatio` isn't _currently_ a ratio, but as the variable name suggests, it should be. The result being that the scaling for `wlRatioRate` goes properly from 0.5 for all losses compared to wins to 1.0 for all wins compared to losses.

Since 0/0 in JS is `NaN`, there's a special case for no trades made where it will set it to 1 (basically 1/1 ratio of wins to losses, even if both were zero).

Also, Euler's number is a built-in mathematic constant in JS, so we can just use it since it's more precise.


DeviaVir (approved):
:+1:

PR #1256, opened 2018-02-01
Fix for 400 bad request from Binance

aquamanjl:
CCXT lib update forced this change (release 1.10.817). Fixes [1254](https://github.com/DeviaVir/zenbot/issues/1254)

### DIFF
--- a/extensions/exchanges/binance/exchange.js
+++ b/extensions/exchanges/binance/exchange.js
@@ -79,7 +79,7 @@ module.exports = function container (get, set, clear) {
       }
 
       var client = publicClient()
-      client.fetchTrades(joinProduct(opts.product_id), '','', args).then(result => {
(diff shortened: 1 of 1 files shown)


Shawn8901:
I think we are also not inserting a correct since value. Can someone check this?

DeviaVir (approved):

aquamanjl:
@Shawn8901 Not sure how the date range is working but it seems to be passed correctly. Here is what I see when running backfill with different days on unstable branch

```
$ ./zenbot.sh backfill binance.ETH-BTC --days 1

binance.ETH-BTC saved 7479 trades 1 days left
..^C
$ ./zenbot.sh backfill binance.ETH-BTC --days 3

binance.ETH-BTC saved 12028 trades 3 days left
^C
$
```

PR #1261, opened 2018-02-01
Various Fixes Multi-Strategy/Correct selections

station384:
Fix multi strategy backtesting
Add overall best result when multi-strategy used
Correct selection when output is displayed as best strategy.
Fix memory leak due to recursive calls. (allows garbage collection of
unused objects)

Make saving of output files synchronous

Change variable assignment of anonymous functions to named functions
(allow easier tracking of entry and exit points)


DeviaVir (approved):
:+1: let's see it in the wild

PR #1263, opened 2018-02-01
Fix ROI

station384:
ROI is used in the status display

This brings back the ROI and fixes the glitch reference in #1258


DeviaVir (approved):
:+1: fair enough

station384:
Concerning ROI,  @glennfu and I submitted updates to darwin within a few minutes of each other.  My update utilized the ROI in the status display,  his removed it completely (valid from his point of view of the code).  

this incorporates a fix in mine that takes into account executing the backtester with capital being 0, It no longer errors.

PR #1265, opened 2018-02-02
update talib to latest release

krystophv:
Puts the `talib` version back to the latest stable release as Windows support has been merged back into the main project. The forked version we're currently on expects to have a bunch of npm tooling that is only default for Node 9.x, so it installs quite a few extra dependencies that we don't need.


DeviaVir (approved):
:+1::+1::+1:

Issue #660, opened 2017-10-27
Strategy onPeriod out of Sequence


KryptoNova:
I'm working with the TALIB functions.  It seems that the execute and callback is being performed on its own thread.  It seems that if the calculation in the TALIB library takes too long, zenbot will move on.  It also appears that the TALIB function calls will queue up and then start to return their results sequentially in their callback functions outside of the main zenbot processing thread.

Here is a screenshot that shows an example:
![image](https://user-images.githubusercontent.com/18106038/32084248-fa8905ec-ba95-11e7-9aaa-c2210779f0bf.png)
If periods were being processed synchronously, then there should be no break between the regular zenbot out and the next number.  These numbers are the result of console.log(result.nbElement) being executed in the TALIB callback inside a function contained in the "lib" directory that is executed in the strategy's onPeriod method.  
Also, without changing the code but simply rerunning, the output shown in the screenshot will not duplicate in the same spot every time.  It seems to be hit or miss where it will happen.

These callbacks seem to be processed asynchronously.  Is there a way to make these TALIB calls process synchronously or is this an illusion from console.log and the standard zenbot output processes?

KryptoNova:
@DeviaVir, I think this may be a larger problem than what I think is possible to correct with a pull request.  Basically, it would force anyone that utilizes "lib" functions to have to rewrite their utilization of it to now process the results of the "lib" call via a callback function.  Perhaps the talib functions can only be used in the "onPeriod" folder so that the "cb" callback can be used to trigger zenbot to move forward.

After doing some tests and reviewing the talib source, this appears to be a problem where the talib functions are executed asynchronously.  This means that the callback from talib may not be ready in time for zenbot's need to calculate it's return values.  I noticed that sometimes i would get sporadic undefined errors, which clued me into this in the first place.

It seems that there may need to be a callback setup from a library function found in the "lib" directory.  That way the talib function can callback to the originating procedure after the calculations are complete.  That way the talib dependency can be satisfied prior to the callback in the "onPeriod" function being executed.

PR #1273, opened 2018-02-03
add PPO (Percentage Price Oscillator) strategy includes rsi oversold trigger

Haehnchen:


KryptoNova:
You will want to move the talib call out of `onCalculate` and instead only use talib functions in the `onPeriod`.  Then use the `onPeriod` `cb` variable as a callback pased into your `lib/ta_ppo`.  Then the callback from the talib execute can in turn utilize the `cb` variable to tell Zenbot the talib is ready for Zenbot to move on.

That  is because talib is processed asynchronously and requires the use of callbacks so Zenbot knows when the talib execute has been completed. The only function that has the ability to andle asynchronous methods in the strategy framework is `onPeriod`.  Otherwise, talib may not finish its calculation before Zenbot has moved on.  

See #660 for more information.

Haehnchen:
@KryptoNova your feedback was very welcome, as i also run into missing calculation. i wrapped the code into promise with a callback on `onPeriod`. indeed a rethink of `onCalculate` would be helpful. will also update #1274 if this solution is accepted

DeviaVir (approved):
:+1:

PR #1277, opened 2018-02-03
add back internal API cache reference for orders for websocket failures

krystophv:
Adds back the "normal" order cache so the API calls have something to grab onto if the websocket connection goes down in the interim.


DeviaVir (approved):
:+1:

Issue #1267, opened 2018-02-02
backfill not working on binance, slow/not working on gdax


dakreepy1:
Please go to Stack Overflow for help and support:

https://stackoverflow.com/questions/tagged/zenbot

If you open a GitHub issue, here is our policy:

1. It must be a bug or a feature request.
2. The form below must be filled out.

**Here's why we have that policy**: Zenbot developers respond to issues. We want to focus on work that benefits the whole community, e.g., fixing bugs and adding features. Support only helps individuals. GitHub also notifies hundreds of people when issues are filed. We want them to see you communicating an interesting problem, rather than being redirected to Stack Overflow.

------------------------

### System information
- **Have I written custom code (as opposed to using zenbot vanilla)**: no
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: ubuntu 16.04
- **Zenbot version**: latest unstable
- **NodeJS version**:  8.9.4
- **Python version**: 2.7.12
- **Exact command to reproduce**:
./zenbot.sh trade binance.ETH-BTC or ./zenbot.sh backfill binance.ETH-BTC
### Describe the problem
after pulling the latest unstable it seems that backfill is broken. There is an error using binance as shown in the the logs section below, seems like the request is missing the startTime parameter.

For gdax ./zenbot.sh backfill gdax.ETH-USD --days 1 i am seeing it get trades but it sure is taking a while, not sure if it was due to the high volume recently or what.  Output shows 100 trades processed after about 3 minutes or so.

When using the trade option (./zenbot trade gdax.BTC-USD) on gdax, the output says 'Backfilling...' and sits there for a while not doing anything, no output at all (waited about 10 minutes).  This was after killing the command and doing  a pull and also an npm update.  So there should not be too many trades to download.

### Source code / logs
ryan@coin:~/zenbot$ ./zenbot.sh backfill binance.ETH-USDT
*************************
Hitting up the exchange 'binance' for trades within the past 7 days.
*************************

Backfilling...

An error occurred { Error: binance GET https://api.binance.com/api/v1/aggTrades?symbol=ETHUSDT&startTime=&endTime=3600000 400 Bad Request {"code":-1105,"msg":"Parameter 'startTime' was was empty."} (possible reasons: invalid API keys, bad or old nonce, exchange is down or offline, on maintenance, DDoS protection, rate-limiting)
[...]

PR #1279, opened 2018-02-03
Refactor backfill for unit testing

haxwell:
* An 'id' and '_id' attribute are set on the resume marker object. This is to match pre-refactor behavior.
* If an exchange uses time as opposed to trade_id in its API queries for trades, the tradeService will get the correct trade from the DB, and pass in its time. #1267


station384:
I was working around this in bittrex it uses a combo of numeric and time depending on where it is getting its backfill data.   I was converting it to numeric only (converting both sources to use ms from 1970)  is this still needed?

station384:
field selector is missing in the trade dataset after backfill refactor.
this field was present in the prior version.

DeviaVir (approved):

PR #1281, opened 2018-02-04
Ensuring startTime and endTime are integers

brinley:
The arguments for startTime and endTime are strings and +360000 appends it as a string breaking the API calls to binance


apekatten:
This fixes Binance for me

DeviaVir (approved):

PR #1298, opened 2018-02-07
eslint fixes and enforcement

krystophv:
Sets up much better eslint integration. Enforces linting on staged files precommit so people can't commit files that don't pass linting.

All the edits to files are so they pass linting. Mostly semicolon or quotes changes. Some unused variables or re-declaration of variables. I tried to leave optional function arguments to preserve the function signature information where relevant. `color` pollutes the `String` native prototype, so it shows up as an unused dependency, which is why there's a bunch of eslint acceptations above them.


DeviaVir (approved):
:+1: _nice_

PR #1302, opened 2018-02-08
Genetic_Backtester: Add launch files for best generation

station384:
added launch parameter --generateLaunch

if testing bittrex.ltc-usdt at the end of each generation backtested
will generate launch files gen.bittrex.ltc-usdt.sh and
gen.bittrex.ltc-usdt.bat these files are configured to launch zenbot
using the best overall settings found for that market during the run.
The files will only be overwritten if the current generation found has
a better fitness than the one in the batch files.

the intention for this feature is two fold. 1st make it easier to use
the results of backtested without the copy pasting.  2nd to allow
backtesting to continuously scan the market testing and evolving with
data that comes in always keeping the launch up to date with the best
set, in another processes, You can launch the file and run it for a set
time at the end of set time it will close and relauch with the new
settings, allowing trading strategy to evolve with backtested findings.
a timeout will need to be added to zenbot trade to support this.


station384:
2 days testing.  working correctly so far.   Found an alternative to modifing zenbot for an exit timeout atleast in linux.

Current Setup
executing all on the same directory,  different terminals.

This scans using 3 strategies and picks the best one, after every generation it creates a file called gen.gdax.bch-btc.sh

Process 1:  Genetic backtesting - terminal 1

`./scripts/genetic_backtester/darwin.js --selector="gdax.bch-btc" --avg_slippage_pct=0.01  --asset_capital="0.0" --currency_capital="0.01434419" --use_strategies="trend_ema,trend_bollinger,trust_distrust"   --population="50" --population_data="" --days="1"  --order_type=maker   --min_periods=15 --generateLaunch`

Next I created a new bash file named **run.gen.bch-btc.sh** with the contents of 
```
#!/bin/bash
while true ; do
timeout --foreground 6h ./gen.gdax.bch-btc.sh
sleep 10
done
```

remember to 
`chmod +x run.gen.bch-btc.sh`

Process 2: zenbot loop - terminal 2
`./run.gen.bch-btc.sh`

Letting this setup run continuously,   the genetic backtest will pick the best strategy for one day in the past,  and every 6 hours zenbot will restart trading using the best strategy found.
So far working good.

DeviaVir (approved):
:+1:

MikeBuzz:
will the above method of running remember previous entry and exit traders when it switches?

the launch file is created with out write permissions so second run fails to update the file

station384:
For the file locks what are you executing  it on?  Linux,  or Mac.
On mac I got got the execute error once till I deleted the file and and corrected the ownership of the file.

To answer the 1st question.   No,  not intentionally.   this does not,  it is only a method of restarting zenbot.   There has been an parameter added to zendbot that allows it to take into account previous trades --use_prev_trades which is not in my sample parameter list as I have it set to default of true in the config file.

MikeBuzz:
Running on Linux, i just changed the permissions on the file and all seems fine, i will check out using that setting, this is a great addition to zenbot, will test with different time periods over the next few days

MikeBuzz:
Permissions on the files are getting reset on every creation
-r----x--t    1 mikebuzz mikebuzz    558 Feb 10 21:05 gen.gdax.btc-eur.bat
-r----x--t    1 mikebuzz mikebuzz    556 Feb 10 21:05 gen.gdax.btc-eur.sh

station384:
The file is updated every cycle (if fitness is >).  I also put in code to make sure it is marked executable.   

There are some things in Node FS that I am not familiar with the full outcomes on different system.   

My expectation and the way I wrote it, was if the file exists, open and truncate it to 0 (erase contents)  write new instructions, and close, set file permissions to executable.

Thes works on my setup (Ubuntu 16 and Ubuntu on Win10 and MacOs).   
 
From the prior messages I thought I saw you were also running ubuntu, and you are getting different results than me.   
Tells me I may have to remove the set executable command.  or go about it in another way, there should be only an access, not a delete and create.

This was the original way I was going to go.

Have a calling script that never changes call the generated script just as a text file.

I decided to go the other way to try and streamline the executing of the cycle without the call to the separate script, seems this may be the way to insure compatibility and then i also do not have to worry about permissions.

station384:
After I posted the last comment, another thought struck me.   


I am running mine on an NFS mapped drive on all my test systems (so its a shared drive, all systems accessing same files).

I’m assuming your running on ext4 local.   I wonder if NFS is masking things on my end allowing mine to work.  I’ll have to look into that.

MikeBuzz:
yes i am using ext4

Burnie09:
Seems the chmod is wrong - at least on *nix.

777 (decimal) == binary 0b 1 100 001 001 == permissions sr----x--x (== octal 1411)
0777 (octal)    == binary 0b 111 111 111    == permissions rwxrwxrwx   (== decimal 511)



```
--- scripts/genetic_backtester/darwin.js.orig	2018-02-11 16:16:21.672459662 +0100
+++ scripts/genetic_backtester/darwin.js	2018-02-11 15:58:30.591932936 +0100
@@ -1050,8 +1050,8 @@
   {
     fs.writeFileSync(lFilenameNix, lNixContents)
     fs.writeFileSync(lFinenamewin32, lWin32Contents)
-    fs.chmodSync(lFilenameNix,777)
-    fs.chmodSync(lFinenamewin32,777)
+    fs.chmodSync(lFilenameNix,0777)
+    fs.chmodSync(lFinenamewin32,0777)
   }
 }
```

edited:
Created a pull request - #1321

LoneWolf345:
Do the simulations ever expire?  For example if darwin finds a really amazing result because the stars aligned and it hit a massive pump and dump event perfectly.  The settings are garbage but because it got lucky it has a high fitness.  Will it just continue to use those settings because nothing in the future will ever beat it?

station384:
What you are asking is a scenario that has occurred to me.   

Currently is just replaces the generated files if the fitness is higher,  so yes that condition could break the pattern.  

I have not been able to solve this.

Some of my ideas vary from removing the fitness check, and just replace the file with the best one after every generation. 
	Problem is darwin will continue to focus on the hump till the analysis range is gone. and you recreate  your gens from the start after you restart darwin

Periodically recreating the generated file no matter if the fitness is more or less. 
	May work.   it does fix the problem where the best fitness in the past no longer applies to current.


I was also trying to fix it by a combination of the 2nd item, and altering the fitness function so it only accepts it as valid if there is more than X many trades.  this will eliminate the large spikes, but then excludes low volume exchanges.

Suggestions would be very helpful, as I’m stuck.  every idea I’ve had there is a greater downside.

my “solution" personally has been to just delete the gen files every day.

LoneWolf345:
if I understand how darwin works, if a result has a top fitness level, it moves on to the next generation.  remove results older than --days= should help as long as days is not terribly long

Primea1:
If you run darwin on 10 plus days to start with you'll likely not be affected by the occasional pump and dump.

adam404:
I have noticed the timeout command in the run.gen.bch-btc.sh script above does not kill the spawned child processes. Which can be problematic. Multiple simultaneous trade bots. Don't have a solution, just wanted to point it out. Tested with mac (gtimeout) and debian (timeout).

station384:
That script was superseded when the –--run_for=<seconds> parameter was added.   

--run_for will not exit till any current trades have completed,  when it exits it exits gracefully.

 

 

My current run script looks like this.

 

#!/bin/bash

 

while true ; do

./gen.gdax.btc-usd.sh --run_for=9200

sleep 10

done

PR #1305, opened 2018-02-09
Remove sosa-mongo, run-parallel dependencies

krystophv:
I wasn't a huge fan of the "magic"/abstraction of the sosa wrapper and after going over the source code, it didn't really do anything except add a couple different layers between the code and the database. It also made it significantly harder to modernize database calls with newer mongo paradigms (read: Promises and async/await). Mostly just need to ad `_id` properties to objects we're saving to the database (because `sosa` just copies the `id` property to `_id`, that's what I did to maintain parity with existing stuff).

The backfill command gets unwound a little, dropping the parallel call in favor of Promise handling and un-nesting functions so we're not re-defining them every loop (read: memory and callstack savings).

I also just noticed that RSI was loading a huge math library to do a really complicated decimal rounding call. Using a native rounding function shaved 15% off of a simulation time.

### DIFF
--- a/commands/backfill.js
+++ b/commands/backfill.js
@@ -1,9 +1,7 @@
 var tb = require('timebucket')
-  , parallel = require('run-parallel')
   , crypto = require('crypto')
-  , tradesCollection = require('../db/trades')
-  , resume_markersCollection = require('../db/resume_markers')
   , objectifySelector = require('../lib/objectify-selector')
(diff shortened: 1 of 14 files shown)


krystophv:
Once again, killing whitespace differences with `?w=1` on the diff view makes it much more digestible. /shrug

DeviaVir (approved):
:+1: less dependencies always +1

Issue #1225, opened 2018-01-25
Question on odd coloring on currency output.


cpjolicoeur:
Curious on the regular expression used for styling the formatted currency here.  It seems to result in some weird currency colors since the regex has 4 different capture groups.  

So `00` shows up one way, but `N0` or `0N` shows up differently (where any non-zero digit in the decimals place is yellow but all zeros are gray.

What was the reasoning behind that out of curiosity instead of just displaying the whole currency string in cyan?

![screenshot](https://i.imgur.com/5mcENIE.png)

cpjolicoeur:
I'd be happy to submit a PR with a change to the coloring to have at least some kind of consistent output regarding the coloring.

trivalik:
Without knowing, just guessing: just highlight the significant numbers, because zero do not matter.

defkev:
Pretty much this.

cpjolicoeur:
It just appears odd to my eyes sometimes having the first digit yellow, sometimes that last digit yellow.

Visuals are totally subjective, of course, so if everyone else likes this styling better, just run with it.  I was mostly curious about the UX design thought process behind it.

Thanks

cpjolicoeur:
I've always seen it displayed like this (with leading zeros the same color until you hit the first non-zero digit).  Even single zeroes would be colored.

![](https://i.imgur.com/Plu1PQn.png)

bitcremental:
In my opinion coloring is not only a matter of _style_, but _meaning_ too (and primarily).
Current coloring scheme for close/current price seems odd to me, because I can't see any meaning in it.
I'd opt for something like this:
1. darkened [trailing zeros](https://en.wikipedia.org/wiki/Trailing_zero), because they don't affect the value (like in the preceding comment);
2. optionally, I'd like to have updated digits highlighted, to catch with a glance the variation respect to last closing price (for example: if last closing price is 123.45 and current price is 124.45, then the characters "4.45" could be highlighted).

What about these criteria? Are there other ones to be considered?
If @cpjolicoeur you have time to work on it and to submit a PR, it's fine for me.

cpjolicoeur:
Sure, I absolutely can submit an update, I'd just want to get some consensus first on what the default coloring/styling of the currency output should be.

Anyone else here care to chime in again?

PR #1311, opened 2018-02-09
Update coloring of formatted currency output

cpjolicoeur:
This is in reference to issue #1225.

This change will "color" up to the first significant digit after the decimal place.  Previously, only non-zero digits were colored.


DeviaVir (approved):
makes sense to me :+1:

PR #1312, opened 2018-02-09
Don't automatically generate html in darwin

glennfu:
With the new folder structure, it's pretty handy to automatically have the html for each sim go into the generation's folder by adding `--include_html`. However, I think people are more likely to be upset to wake up in the morning and find over 1GB of space has been taken up by their experiment. Let's leave this off by default.


station384:
Agree.  with my pull/feature I submitted it would be running 24/7.   that is alot of wasted space.

DeviaVir (approved):

PR #1313, opened 2018-02-10
Changed all tests to use actual, and consistent, millisecond times

haxwell:
All the trades used in test now have consistent IDs and timestamps. #sanity


DeviaVir (approved):
nice!

PR #1321, opened 2018-02-11
Fix file permissions on launch files

Burnie09:
Need to be octal 0700, not decimal 777

Also fixed date formatting ( hh - 00-12   HH - 00-24 )

### DIFF
--- a/scripts/genetic_backtester/darwin.js
+++ b/scripts/genetic_backtester/darwin.js
@@ -55,13 +55,13 @@ let darwinMonitor = {
   actualRange: function(so) {
     // Adapted from sim.js logic to similarly figure out how much time is being processed
     if (so.start) {
-      so.start = moment(so.start, 'YYYYMMDDhhmm')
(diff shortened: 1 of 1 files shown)


station384:
Ty. I missed the octal decimal bit ( sadly I usually do as I flip between *nix and windows, JS and java and .net.. )

station384:
@Burnie09 I'm going to incorporate your fixs into mine,  I have a few other changes to add to this.
I for the octet I used the string version '777'   so eslint won't complain about "invalid number".   according to the FS source code the mode can be passed in as a number (octet) or string.  so 0777 would be '777' as a string.  

I am also changing the launch files so they start the node process directly instead of a bash shell.  I've noticed that after a few cycles timeout doesn't kill the process completely if its embedded to many times  i.e. timeout->bash->bash      last one can linger if the kill signal gets lost.  

I am also adding a run_for parameter to the trade command to allow for clean exits of zenbot without using the timeout command in linux.

DeviaVir (approved):

PR #1322, opened 2018-02-12
Genetic_backtester

station384:
incorporates changes in pull #1321
Changed launch files so they execute node directly and not thru bash.
this fixes an orphaned process problem i was seeing where the last
launched process was not getting the kill signal.

Add parameter —run_for to trade command.  this accepts an integer, and
represents how many minutes to trade before doing a clean exit.

### DIFF
--- a/commands/trade.js
+++ b/commands/trade.js
@@ -49,11 +49,13 @@ module.exports = function (program, conf) {
     .option('--use_prev_trades', 'load and use previous trades for stop-order triggers and loss protection')
     .option('--disable_stats', 'disable printing order stats')
     .option('--reset_profit', 'start new profit calculation from 0')
+    .option('--run_for <minutes>', 'Execute for a period of minutes then exit with status 0', String, null)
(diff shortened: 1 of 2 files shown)


DeviaVir (approved):

narentin:
If `--run-for` parameter is not given, could we the previous behavior (ie. run forever)?

narentin:
When `--run-for` parameter is used, the following warning appears:
Deprecation warning: moment().add(period, number) is deprecated. Please use moment().add(number, period). See http://momentjs.com/guides/#/warnings/add-inverted-param/ for more info.

The following removes the warning:
`var botStartTime = moment().add(so.run_for,'m')`

illya13:
@station384 @DeviaVir 
+1 request to keep previous behavior if no `--run-for` or provide option to have it like `-1`

station384:
@narentin bug fixed by @illya13 in pull #1329

PR #1323, opened 2018-02-12
Enable GetorderBook on Bittrex and Poloniex

ljankok:
Enable GetorderBook on Bittrex and Poloniex and provide example in strategy.js of noob

### DIFF
--- a/extensions/exchanges/bittrex/exchange.js
+++ b/extensions/exchanges/bittrex/exchange.js
@@ -273,6 +273,38 @@ module.exports = function bittrex(conf) {
       })
     },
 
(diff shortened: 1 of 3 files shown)


Shawn8901:
Hi i like this in general, but wouldnt it be more zenbot'ish to do stuff like getting the order book in the engine and store it in a variable like s.period.orderBook?

ljankok:
I did it like this because the fetch result of the order book is different per exchange.

DeviaVir (approved):
:+1: good extension on these exchanges, how we'll implement these is up to future contributors

PR #1324, opened 2018-02-12
Intro "global" eventEmitter for loosely coupled inter-module communication

krystophv:
This PR introduces a concept (or [architecture](https://en.wikipedia.org/wiki/Event-driven_architecture)) I'd like to gradually develop into a main paradigm within the app of a more or less globally available (i.e. on the `conf` object that gets passed to pretty much every module) [EventEmitter](https://nodejs.org/api/events.html#events_class_eventemitter) that will allow modules to subscribe to specific types of events and publish specific types of events as necessary. The [pub/sub](https://en.wikipedia.org/wiki/Publish–subscribe_pattern) style is a well established paradigm that works well for apps with a sort of "plugin" style architecture that Zenbot currently leans toward (for things like `output` and `notifiers` specifically, but even `strategies` and `exchanges` can fall into that category too). 

In this PR, I've only really moved the `sim` command over as a beginning point, with some modifications to `engine` to support it. Instead of `engine` repeatedly polling for sets of trades from `sim`, now it can simply set a handler for `trade` events from the global eventBus and `sim` can `emit` `trade`s on the global eventBus. Neither of them need to call each other's functions directly. Since this takes out the repeat call/callback model, `sim` is free to publish events to the bus directly from the database stream instead of buffering into an `Array` first, letting us remove our query limit and allowing for much better overall memory consumption and garbage collection by the runtime (all references to `trade` are dead almost immediately after `onTrade`). 

Unraveling the `update` function in `engine` led to a clear opportunity to get rid of the `run-series` call which I believe was leading to some maximum call stack issues for some people since each trade callback would execute the next. I left the function signature as is for now until more pieces of the app (specifically `trade`) get updated to use the eventBus. 

I also took a little time to clean up the boot/index/zenbot.js situation which was doing a handful of things that were either unnecessary or not useful. `sim` was also trying to take a second stab at overrides from `conf` where the command line arguments should have the final override.

And finally, introduces a `--silent` option for `sim` which simply doesn't output anything until the sim has completed. This can significantly speed up the execution time of the sim (up to %50+ for small period lengths that have a lot of period reports) and can be very useful in `darwin` where the output isn't super useful. Using this option does have the side effect of not getting all the nice reporting from the new darwin progress interface, but I prefer the speed over the stats and it's optional, so people can pick.

As always, totally open to feedback/discussion.

### DIFF
--- a/commands/sim.js
+++ b/commands/sim.js
@@ -47,6 +47,12 @@ module.exports = function (program, conf) {
       var s = { options: minimist(process.argv) }
       var so = s.options
       delete so._
+      if (cmd.conf) { 
+        var overrides = require(path.resolve(process.cwd(), cmd.conf)) 
+        Object.keys(overrides).forEach(function (k) { 
+          so[k] = overrides[k] 
+        }) 
+      }
       Object.keys(conf).forEach(function (k) {
         if (!_.isUndefined(cmd[k])) {
           so[k] = cmd[k]
--- a/commands/sim.js
+++ b/commands/sim.js
@@ -42,6 +42,7 @@ module.exports = function (program, conf) {
     .option('--enable_stats', 'enable printing order stats')
     .option('--backtester_generation <generation>','creates a json file in simulations with the generation number', Number, -1)
     .option('--verbose', 'print status lines on every period')
+    .option('--silent', 'only output on completion (can speed up sim)')
     .action(function (selector, cmd) {
       var s = { options: minimist(process.argv) }
       var so = s.options
--- a/commands/sim.js
+++ b/commands/sim.js
@@ -8,6 +8,7 @@ var tb = require('timebucket')
   , objectifySelector = require('../lib/objectify-selector')
   , engineFactory = require('../lib/engine')
   , collectionService = require('../lib/services/collection-service')
+  , _ = require('lodash')
 
 module.exports = function (program, conf) {
   program
@@ -42,15 +43,16 @@ module.exports = function (program, conf) {
     .option('--backtester_generation <generation>','creates a json file in simulations with the generation number', Number, -1)
     .option('--verbose', 'print status lines on every period')
     .action(function (selector, cmd) {
-      var s = {options: minimist(process.argv)}
+      var s = { options: minimist(process.argv) }
       var so = s.options
       delete so._
       Object.keys(conf).forEach(function (k) {
-        if (typeof cmd[k] !== 'undefined') {
+        if (!_.isUndefined(cmd[k])) {
           so[k] = cmd[k]
         }
       })
       var tradesCollection = collectionService(conf).getTrades()
+      var eventBus = conf.eventBus
 
       if (so.start) {
         so.start = moment(so.start, 'YYYYMMDDhhmm').valueOf()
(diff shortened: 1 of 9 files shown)


commands/sim.js
```diff
@@ -41,16 +42,18 @@ module.exports = function (program, conf) {
     .option('--enable_stats', 'enable printing order stats')
     .option('--backtester_generation <generation>','creates a json file in simulations with the generation number', Number, -1)
     .option('--verbose', 'print status lines on every period')
+    .option('--silent', 'only output on completion (can speed up sim)')
```
DeviaVir, line 45: I'm thinking `--silent` and `--verbose` are slightly duplicate, in the sense that the current output with all of the lines showing exactly what and when the strategy does things is more useful for strategy developers (and thus should be considered verbose) where a consumer would only be interested in sim results.
What do you think about dropping `--silent` and moving showing output lines only if `--verbose` is set?

lib/engine.js
```diff
@@ -835,6 +841,88 @@ module.exports = function (s, conf) {
...
+        if (s.signal) {
+          executeSignal(s.signal)
+          s.signal = null
+        }
+      }
+      if (so.mode !== 'live') {
+        adjustBid(trade)
+        executeOrder(trade)
+      }
+    }
```
DeviaVir, line 868: (out of scope for this PR, I think, but:) I am wondering if we should have a function here for stop loss/profit stop, since I'm guessing we're only doing those on periods now, where it might make more sense to do it on trades. Thoughts?

DeviaVir (approved):
Absolutely like what you did with the EventEmitter and look forward to seeing that implemented everywhere.

commands/sim.js
```diff
@@ -77,12 +80,6 @@ module.exports = function (program, conf) {
       so.selector = objectifySelector(selector || conf.selector)
       so.mode = 'sim'
 
-      if (cmd.conf) {
```
krystophv: I need to put this back in, actually, but in my opinion, above the bits for the CLI args so things passed in on the command line get the ultimate override. And/or we could move more things to stop using `options` and use the `conf` since it should contain a proper set of configuration, but that's going to be a change outside the scope for this PR.

commands/sim.js
krystophv: So they currently do different things. The `--verbose` option turns on the status lines for sim and train that have a carriage return at the end, generating the tape of new entries on every period. The `--silent` option actually turns off the `stdout` output where you'll see the system overwriting the same line over and over again with the current status line (the bit thats always flashing during live/sim). I didn't want to have it default to one or the other as the nice new darwin output actually reads that stdout output in order to provide progress for individual sims, so the new hotness doesn't look quite as nice (it still gives number of active sims, just not an estimate on time).

lib/engine.js
krystophv: I'd tend to agree. We'll have to approach it a little carefully as a single trade outside the thresholds could trigger the system to initialize a stop order where currently the period has to close outside the threshold. I'd like the stop analysis to at least happen onTrade and then could emit a buy/sell signal on the eventBus independent of other pieces of the system. Having buy/sell signals come over the bus also means that strategies could also emit them from their `calculate` functions if they wanted to.

DeviaVir:
I'll merge this later today if nobody objects.

PR #1330, opened 2018-02-12
Genetic_Backtester

station384:
Allow passing of extra parameters into generated scripts 

change chmodsync from octet to string to keep .eslint from complaining
about “invalid number”

### DIFF
--- a/scripts/genetic_backtester/darwin.js
+++ b/scripts/genetic_backtester/darwin.js
@@ -1045,18 +1045,18 @@ function saveLaunchFiles(saveLauchFile, configuration ){
   //write Nix Version
   let lNixContents = '#!/bin/bash\n'.concat('#fitness=',configuration.fitness,'\n',
     'env node zenbot.js trade ', 
-    bestOverallCommand,'\n')
+    bestOverallCommand,' $@\n')
   let lWin32Contents = '@echo off\n'.concat('rem fitness=',configuration.fitness,'\n',
     'node zenbot.js trade ', 
-    bestOverallCommand,'\n')
+    bestOverallCommand,' %*\n')
   
   if (Number(configuration.fitness) > Number(lastFitnessLevel))
   {
     fs.writeFileSync(lFilenameNix, lNixContents)
     fs.writeFileSync(lFinenamewin32, lWin32Contents)
     // using the string instead of octet as eslint compaines about an invalid number if the number starts with 0
-    fs.chmodSync(lFilenameNix, 0777)
-    fs.chmodSync(lFinenamewin32, 0777)
+    fs.chmodSync(lFilenameNix, '777')
+    fs.chmodSync(lFinenamewin32, '777')
   }
 }
 
--- a/scripts/genetic_backtester/darwin.js
+++ b/scripts/genetic_backtester/darwin.js
@@ -55,13 +55,13 @@ let darwinMonitor = {
   actualRange: function(so) {
     // Adapted from sim.js logic to similarly figure out how much time is being processed
     if (so.start) {
-      so.start = moment(so.start, 'YYYYMMDDhhmm')
+      so.start = moment(so.start, 'YYYYMMDDHHmm')
       if (so.days && !so.end) {
         so.end = so.start.clone().add(so.days, 'days')
       }
     }
     if (so.end) {
-      so.end = moment(so.end, 'YYYYMMDDhhmm')
+      so.end = moment(so.end, 'YYYYMMDDHHmm')
       if (so.days && !so.start) {
         so.start = so.end.clone().subtract(so.days, 'days')
       }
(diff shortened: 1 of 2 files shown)


scripts/genetic_backtester/darwin.js
```diff
@@ -1045,18 +1045,18 @@ function saveLaunchFiles(saveLauchFile, configuration ){
...
   
   if (Number(configuration.fitness) > Number(lastFitnessLevel))
   {
     fs.writeFileSync(lFilenameNix, lNixContents)
     fs.writeFileSync(lFinenamewin32, lWin32Contents)
     // using the string instead of octet as eslint compaines about an invalid number if the number starts with 0
-    fs.chmodSync(lFilenameNix, 0777)
-    fs.chmodSync(lFinenamewin32, 0777)
+    fs.chmodSync(lFilenameNix, '777')
+    fs.chmodSync(lFinenamewin32, '777')
```
DeviaVir, line 1059: not `'0777'`?

station384:
According to all the docs I’ve read and checking the FS source,  it takes 2 types of input.   a number represented as octets (0777 the zero in front designating it as an octet number) or a string.  the string being a literal representation of the octet value of 777.
it sends the string thru a conversion, which results in 0777 number or 511 decimal or 1FF hex, depending on how you view it.
I would have left it as 0777 since it eliminates the internal conversion but .eslint complains that 0777 is an “invalid number” and my post checks eslint errors and prevents posts if there are any, and since this isn’t a tight loop area, the extra conversion doesn’t hurt, so decided on the string route.
this link started my research on it once I found out I made the mistake, and why it is the way it is.
https://stackoverflow.com/questions/8756639/how-do-i-use-chmod-with-node-js <https://stackoverflow.com/questions/8756639/how-do-i-use-chmod-with-node-js>
from there it went to the FS source and tracking it on thru.

scripts/genetic_backtester/darwin.js
station384: According to all the docs I’ve read and checking the FS source, it takes 2 types of input. a number represented as octets (0777 the zero in front designating it as an octet number) or a string. 
the string being a literal representation of the octet value of 777.  it sends the string thru a conversion, which results in 0777 number or 511 decimal or 1FF hex, depending on how you view it. 

I would have left it as 0777 since it eliminates the internal conversion but .eslint complains that 0777 is an “invalid number” and my post checks eslint errors and prevents posts if there are any, and since this isn’t a tight loop area, the extra conversion doesn’t hurt, so decided on the string route. 

This link started my research on it once I found out I made the mistake, and why it is the way it is. https://stackoverflow.com/questions/8756639/how-do-i-use-chmod-with-node-js  from there it went to the FS source and tracking it on thru.

krystophv: Yeah, that's a deprecated octal format. It seems like there's a couple solutions, but the current changes likely work. ref: https://github.com/babel/babel/issues/2569

DeviaVir (approved):

scripts/genetic_backtester/darwin.js
station384: Ty.  I didn't realize about the 0o designation.

phatwila:
@DeviaVir 
The just merged changes cause the script to hang, which were not an issue right before I pulled them down:
```

[root@cpaneldev zenbot]# scripts/genetic_backtester/darwin.js --selector="gdax.BTC-USD" --use_strategies="tma" --days=10 --asset_capital=0.001 --currency_capital=1400 --days=10

--==Zenbot 4 Genetic Backtester v0.2.2==--
Tue, 13 Feb 2018 22:03:24 GMT

Backtesting strategy tma ...

Creating population of 100 ...



=== Simulating generation 0 ===

Backfilling (might take some time) ...
./zenbot.sh backfill --days=10,10 gdax.BTC-USD

```

station384:
Days is entered 2x
And Gdax can take a long time to backfill. The days entered 2x causes it to add them together result in a backfill of days=10,10 and will cause it to backfill from 1970 till current. So it will look like it hung.  

On phone right now so I can’t check but I don’t think Tma strategy is not supported.

phatwila:
@station384 
Oops - Well I took out the extra date, and switched to dema (even though my TMA strategy functioned properly before the pull) and its still hanging. GDAX has already been back filled before this.  checking top, it appears the sim is running, just no screen output. Any other details I can provide?

station384:
okay.   looks good.   I would bet coin that it is gdax backfilling that is taking so long (I’ve had it take 5hrs before for backfilling 15 days)

gdax is special,  it uses websockets, so it responds when the server sends an update, not when we request itso we get it at the rate gdax determines, and when its busy it can come in VERY slowly.

When darwin is backfilling it does not give a status of where its at.     

If I’m doing a large backfill I will prefill using the command manually so I can see a status and get an idea if it has hung or not.  the command would be

./zenbot.sh backfill gdax.BTC-USD —days=10

after that finishes,  rerun darwin.js  is should just backfill only the data since the last backfill.

phatwila:
@station384 
Yes, I actually ran a 10 day backfill directly before executing the command so data was upto date (painfully slow as usual with GDAX). After about 10 minutes of letting it sit there, it finally began to run, but each generation is taking close to 25 minutes to complete
Generation 3 completed at 2018-02-14 12:12:14, took 25m 38s, results saved to

where as before the pull generations with the same data settings were completing in minutes.. Not really sure what else to check.

krystophv:
I'd recommend trying out the `--silent` option and see if that improves anything for you. Other than that, if you're running on a machine that you have GUI access for, then you can run an individual sim through a debugger via `node --inspect-brk zenbot sim <args>` and then open Chrome `about:inspect` and open the dedicated Node debugging tools, go to the `Profiler` tab and start a profile, then end it once the sim has completed and it'll let you know where it's spending it's time.

PR #1331, opened 2018-02-12
add talib "ULTOSC - Ultimate Oscillator" strategy with oversold

Haehnchen:


Haehnchen:
need some more validation first

PR #1338, opened 2018-02-14
only instantiate one copy of an exchange

krystophv:
This just makes it so that `trade` isn't making it's own instantiated copy of an `exchange` and letting `engine` make another. Instead, it will instantiate one and attach it to `s`, which `engine` can check for before instantiating it's own. `engine` does need to instantiate one for all the other commands that call for it, so it maintains that ability in the case that it hasn't been done already.


DeviaVir (approved):
:+1:

PR #1339, opened 2018-02-14
Fix strategy Noop

station384:
noop now does nothing instead of throwing an error.


DeviaVir (approved):

ljankok (commented):
this code is an example for poloniex and bittrex only

PR #1345, opened 2018-02-15
Faster Docker build

jerson:
avoid downloading the dependencies in each build, they are downloaded only if the package-lock.json or package.json is modified

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,9 +3,10 @@ FROM node:8
 RUN mkdir -p /app
 WORKDIR /app
 
-COPY . /app
 RUN npm install -g node-gyp
+COPY package-lock.json package.json post_install.js ./
 RUN npm install --unsafe
+COPY . .
 
 RUN ln -s /app/zenbot.sh /usr/local/bin/zenbot
 


Dockerfile
```diff
@@ -3,9 +3,10 @@ FROM node:8
 RUN mkdir -p /app
 WORKDIR /app
 
-COPY . /app
 RUN npm install -g node-gyp
+COPY package-lock.json package.json post_install.js ./
 RUN npm install --unsafe
+COPY . .
```
DeviaVir, line 9: You're copying to `./` but the below run is still linking to `/app`, either keep `/app` or update everything to `/`

jerson: hi @DeviaVir, when you define WORKDIR, that is the base directory and it is easier to use only "." that "/app" every time

note: try to compile the dockerfile once, that will take the normal time, then if you modify a javascript file and recompile, it will continue only from line 9 and it will be much faster

DeviaVir: Ah yes, you're right about the `WORKDIR`. However, the `post_install.js` runs some things that are not copied yet (e.g. `webpack` sources and `scripts/genetic_backtester`): https://github.com/DeviaVir/zenbot/blob/unstable/post_install.js#L8

I think this is why I decided it might make more sense to copy everything and then install. Do you see a way around that?

jerson: you're right, forget that, the solution would be to organize the code in another way, maybe moving the script code to your own packages npm would solve it

DeviaVir: Agreed, maybe even as completely separate containers...

PR #1327, opened 2018-02-12
conf override fixes

krystophv:
More robust command line input for conf override and prevention of modifying a `require` cached version of the conf by making a clone of it for the app runtime.

Should fix #1304

PR #1347, opened 2018-02-15
Reorder and fix config file loading

defkev:
This is a rework of #1327

Fixes:
* If conf.js is missing fall back to conf-sample as "default" config
* --~~conf oddly looking for the config file in the selector directory~~ this is actually just a display bug because the error message uses `__dirname` instead of `process.cwd()` the real problem gets fixed by above

Config files will now get loaded in this order on boot:

1. conf overrides file if provided by `--conf`
2. `conf.js` if present
3. and finally `conf-sample.js` to fill up possible blanks

Priority of the settings (from left to right) are `--conf -> conf.js -> conf-sample.js`

### DIFF
--- a/boot.js
+++ b/boot.js
@@ -8,16 +8,9 @@ module.exports = function (cb) {
   var zenbot = { version }
   var args = minimist(process.argv.slice(3))
   var conf = {}
+  var config
 
-  // 1. load default config
-  try {
-    defaults = require('./conf')
-  } catch (err) {
-    console.error(err + ', falling back to conf-sample.js')
-    defaults = require('./conf-sample')
-  }
-
-  // 2. load custom config
+  // 1. load conf overrides file if present
   if(!_.isUndefined(args.conf)){
     try {
       conf = require(path.resolve(process.cwd(), args.conf))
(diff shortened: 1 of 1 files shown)


boot.js
```diff
@@ -7,20 +7,26 @@ var EventEmitter = require('events')
 module.exports = function (cb) {
   var zenbot = { version }
   var args = minimist(process.argv.slice(3))
-  var conf
+  var conf = {}
 
+  // 1. load default config
+  try {
+    defaults = require('./conf')
```
krystophv: If `conf` isn't a _full_ copy of the `conf-sample` (for instance, if it only holds the values that a user wanted to customize), then it doesn't suffice as a full set of values for `defaults`, but the `conf-sample` will never be loaded.

defkev: Indeed, thats why we tell people to make a copy of `conf-sample.js` as `conf.js` in the README
> cp conf-sample.js conf.js

My personal best practice usually goes like this:

1. Copy `conf-sample.js` to `conf.js`
2. Change `conf.js` (API keys, notifiers, etc.)
3. Create a custom conf (e.g. `cex_btc_usd.conf`) with period, neutral_rate, etc. for each instance to load with `--conf=...`

This way i have one base config and multiple custom configs to cover my separate instances on a single node.

IMHO sometimes its simpler to get people to "read (and follow) the f-ing manual" 😁 instead of trying to cover every possible constellation they might come up with.

krystophv: That would work if we insist that users also update their `conf.js` every time a new option is added and a new default value is added to `conf-sample.js`, but I think that's somewhat unreasonable to ask.

evseevnn: Yes, i think what better just merge it with default

defkev: Fair point if the committer forgets to check if his/her new property actually exists.

> but I think that's somewhat unreasonable to ask

Not for me 😁 

Let me rework it.
brb

defkev: Done.

I updated the merge comment to reflect the changes.

DeviaVir (approved):

PR #1357, opened 2018-02-17
add new indicator for talib "WILLR - Williams' %R with rsi oversold", "ADOSC - Chaikin A/D Oscillator" with rsi oversold with support for all volume based indicator

Haehnchen:
some more indicator, still no strategy because they need some combinations for a suitable outcome

### DIFF
--- a/lib/ta_volume.js
+++ b/lib/ta_volume.js
@@ -1,0 +1,62 @@
+// allows to use all talib Volume Indicator Functions: https://mrjbq7.github.io/ta-lib/func_groups/volume_indicators.html
+// AD - Chaikin A/D Line
+// ADOSC - Chaikin A/D Oscillator
+// OBV - On Balance Volume
+
+var talib = require('talib')
+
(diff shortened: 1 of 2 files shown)


lib/ta_volume.js
```diff
@@ -0,0 +1,62 @@
...
+        s.marketData.volume.push(s.lookback[i].volume)
+      }
+    }
+
+    if (s.marketData.close.length < min_periods) {
+      resolve()
+      return
+    }
+
+    let tmpHigh = JSON.parse(JSON.stringify(s.marketData.high))
```
krystophv: The JSON stringify/parse is an incredibly slow way to make an array copy. Please consider `array.slice(0)` instead. [array.slice() documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice) Benchmark: http://jsben.ch/dzO4U

Haehnchen: thx for this input done

DeviaVir (approved):
:+1:

PR #1365, opened 2018-02-17
prevent modification of require() cache for overrides

krystophv:
This modifies the `_.defaultsDeep` call to extend a local empty object instead of an object that comes from a `reauire()` cache. 

Also specifies that we get all the mongo config values from the `zenbot.conf` copy.

### DIFF
--- a/boot.js
+++ b/boot.js
@@ -9,11 +9,12 @@ module.exports = function (cb) {
   var args = minimist(process.argv.slice(3))
   var conf = {}
   var config = {}
+  var overrides = {}
 
   // 1. load conf overrides file if present
   if(!_.isUndefined(args.conf)){
     try {
-      conf = require(path.resolve(process.cwd(), args.conf))
+      overrides = require(path.resolve(process.cwd(), args.conf))
     } catch (err) {
       console.error(err + ', failed to load conf overrides file!')
     }
@@ -21,16 +22,15 @@ module.exports = function (cb) {
 
   // 2. load conf.js if present
   try {
-    config = require('./conf')
+    conf = require('./conf')
   } catch (err) {
     console.error(err + ', falling back to conf-sample')
   }
 
   // 3. Load conf-sample.js and merge
   var defaults = require('./conf-sample')
-  _.assign(config, conf)
-  _.defaultsDeep(config, defaults)
-  zenbot.conf = _.cloneDeep(config)
+  _.defaultsDeep(config, overrides, conf, defaults)
+  zenbot.conf = config
 
   var eventBus = new EventEmitter()
   zenbot.conf.eventBus = eventBus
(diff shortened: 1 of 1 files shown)


boot.js
```diff
@@ -28,30 +28,31 @@ module.exports = function (cb) {
 
   // 3. Load conf-sample.js and merge
   var defaults = require('./conf-sample')
-  _.defaultsDeep(conf, config, defaults)
```
defkev, line 31: imho redundant as `--conf` already takes precedence over `conf.js`

1. `conf` gets filled by `--conf` (the object)
2. `config` (content of `conf.js`) and `defaults` (content of `conf-sample.js`) sources fill `conf` with all new properties

The priority should always be `--conf -> conf.js -> conf-sample.js` 

We call `--conf` config **overrides** for a reason.

defkev (changes requested):
> if a user has a conf.js with some general overrides defined but also provides a --conf configuration file override, then _.defaultsDeep will only copy values from the --conf file where conf.js has left undefined

Wrong

e.g. take `c.strategy = 'trend_ema'` in `conf.js` then override it using `--conf=whatever.conf` (lets say by `c.strategy = 'rsi'`) `c.strategy` will be `rsi` because it is already present in the `object` of `defaultsDeep`

`--conf` is the `object` not a `source`

See https://lodash.com/docs/4.17.5#defaultsDeep

boot.js
krystophv: Indeed, you're correct. I misdiagnosed a different issue for this. However, I'm still willing to submit this as-is because `_.defaultsDeep()` mutates the source object, which in this case is the object from the `require()` cache and I think it's a bad practice to modify that object since it gets `require()`d elsewhere in the app and that code expects it to be unmodified.

defkev: I initially had `_.defaultsDeep(conf, ...` as an empty object and sourced it using `overrides, config, defaults)` in yesterdays PR but reworked it as it looked like an overkill.

> mutates the source object

According to the docs this only mutates the object (not the sources) which i am fine this as this is indeed our running config. The require alone wouldn't suffice anyway, unless someone parses an entire working config using `--conf` of course.

krystophv: By mutating the object (which is what is going to be returned on subsequent `require()` calls to the override path), we end up providing a full copy of the config to the commands that process the overrides for their internal `s.options` i.e.: https://github.com/DeviaVir/zenbot/blob/unstable/commands/sim.js#L52 - instead of just the subset in the overrides file, which will be a much larger set of properties than it currently retains.

krystophv: I actually prefer the idea of what you described as your initial implementation, using an empty object as the source. I might update this to just do that.

defkev (approved):
👍

DeviaVir (approved):

PR #1375, opened 2018-02-19
Actually use avg_slippage_pct to adjust balances

krystophv:
Discovered a bug where currently the `balance.currency` get's adjusted with a `price` _before_ that `price` is adjusted with potential `avg_slippage_pct`. Also, the `fee` is being calculated off the `price` before applying the `avg_slippage_pct`. In it's current state, `avg_slippage_pct` has no effect on the running and final balance - only the price that gets recorded in an individual trade (in the `my_trades`) array, but that's not used in any of the final calculations (i.e. `vsBuyHold`, `profit`, etc) - only the `balance.currency` is. Once taking the `avg_slippage_pct` into account for the `balance.currency`, simulations seem to, in general, perform worse (and most likely more in-line with reality).


Shawn8901 (commented):
Nice job 👌

DeviaVir (approved):

PR #1378, opened 2018-02-20
Implement sim exchange to make sim perform more like live

glennfu:
I would like approval from @DeviaVir @defkev and @krystophv before this one gets accepted, mostly to make sure I didn't break any of their recent changes.

Fair warning, this WILL change sim results from what people are used to and that will probably scare them, but that's the point. Plenty of people complain that sim vs live is quite different, so I set out to close that gap as much as possible.

Current unstable sim:
sim

My output from live:
live

Sim with my PR in place:
sim2

You can see the new version is MUCH more in line with reality. The way I did this was by implementing a "sim" exchange. I moved as much `if (mode === 'live')` portions of engine.js as I could and abstracted them away into the sim exchange.

I also implemented an `eventQueue` to mimic the effect of `setTimeout` in a way that would translate to sim, so that we can get more similar results to the `conf.wait_for_settlement` and `so.order_poll_time` settings that live uses.

Next steps:

1. Make the sim exchange do something with an abstraction of the `eventQueue` so that we can simulate network delay (like 150ms?) on its commands.
2. Make the sim exchange compare not only trade price, but trade size, in order to allow sim to process partial orders. Right now partial orders in live totally screw me up, but I can't test them in sim. That needs to be possible.
3. Figure out why my candles from live don't look exactly like my candles from sim. I have no good leads there, I just know it's a problem and I can't figure out why.

### DIFF
--- a/lib/engine.js
+++ b/lib/engine.js
@@ -21,7 +21,12 @@ module.exports = function (s, conf) {
 
   let so = s.options
   if(_.isUndefined(s.exchange)){
-    s.exchange = require(path.resolve(__dirname, `../extensions/exchanges/${so.selector.exchange_id}/exchange`))(conf)
+    if (so.mode == 'sim') {
+      s.exchange = require(path.resolve(__dirname, '../extensions/exchanges/sim/exchange'))(conf, s)
+    }
+    else {
+      s.exchange = require(path.resolve(__dirname, `../extensions/exchanges/${so.selector.exchange_id}/exchange`))(conf)
+    }
   }
   s.product_id = so.selector.product_id
   s.asset = so.selector.asset
@@ -43,7 +48,7 @@ module.exports = function (s, conf) {
     console.error('error: could not find product "' + s.product_id + '"')
     process.exit(1)
   }
-  if ((so.mode === 'live' || so.mode === 'paper') && s.exchange.dynamicFees) {
+  if (s.exchange.dynamicFees) {
     s.exchange.setFees({asset: s.asset, currency: s.currency})
   }
   if (so.mode === 'sim' || so.mode === 'paper') {
@@ -87,11 +92,21 @@ module.exports = function (s, conf) {
     if (s.strategy.getOptions) {
       s.strategy.getOptions.call(s.ctx, s)
     }
+    if (s.strategy.orderExecuted) {
+      eventBus.on('orderExecuted', function(type) {
+        s.strategy.orderExecuted(s, type, executeSignal)
+      })
+    }
   }
 
   function msg (str) {
     if (so.debug) {
-      console.error('\n' + moment().format('YYYY-MM-DD HH:mm:ss') + ' - ' + str)
+      if (lastTickCheckOrder) {
+        console.error('\n' + moment(lastTickCheckOrder).format('YYYY-MM-DD HH:mm:ss') + ' - ' + str)
+      }
+      else {
+        console.error('\n' + moment().format('YYYY-MM-DD HH:mm:ss') + ' - ' + str)
+      }
     }
   }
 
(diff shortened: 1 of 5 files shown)


Shawn8901:
Hi, i definetly like the idea. I found some points which should be fixed: you commited a strategy with that commit. If you are willing to share that, i would prefere it as an extra pull request, instead of a "hidden" commit here. Same with the indicator.  Also you commited "dead" code (in darwin.js) which relates to other  uncommitted strategies. You should remove those things. 
Also the engine backup is not needed. Git has everything for restoring the old version on board.
(i did yet no functional testing)

glennfu:
Totally didn't mean to include those extra files. Thanks for pointing that out. I removed those extraneous changes.

lib/engine.js
```diff
@@ -872,19 +759,157 @@ module.exports = function (s, conf) {
...
+      if (now() - order.local_time >= so.order_adjust_time) {
+        getQuote(function (err, quote) {
+          if (err) {
+            err.desc = 'could not execute ' + type + ': error fetching quote'
+            return cb(err)
+          }
+          let marked_price
+          if (type === 'buy') {
+            marked_price = nextBuyForQuote(s, quote)
+            if (so.exact_buy_orders && n(order.price).value() != marked_price) {
```
defkev, line 882: What is `so.exact_buy|sell_orders` as it never seems to get populated?

lib/engine.js
```diff
@@ -21,8 +21,13 @@ module.exports = function (s, conf) {
 
   let so = s.options
   if(_.isUndefined(s.exchange)){
+    if (so.mode == 'sim') {
+      s.exchange = require(path.resolve(__dirname, '../extensions/exchanges/sim/exchange'))(conf, s)
+    }
+    else {
     s.exchange = require(path.resolve(__dirname, `../extensions/exchanges/${so.selector.exchange_id}/exchange`))(conf)
   }
```
DeviaVir: weird indenting here

glennfu: Thanks

lib/engine.js
glennfu: I wrote this as an additional strategy option. I've been using it in live for the last couple of weeks. The default `if (n(order.price).value() > marked_price)` is horrible for the strategy I'm running. It's very important to me that the price get adjusted in both directions so I add `--exact_buy_orders` to my command to solve that.

defkev: > that the price get adjusted in both directions

Why no PR for this?

This might be useful to others as well and frankly, it makes little sense to have code in upstream that never gets reached.

glennfu: It's been a bit tricky trying to isolate all the things I'm working on at the same time. The only way I have to test this code is in live, because we don't have a good enough test framework in place already. So to pull these things out separately would be to break my bot, and then I don't have a way to test that what I'm doing is working or not. I agree with you in theory, but it's too hard in practice. Luckily it's written an as entirely optional feature that by default won't change anything. I'll go back and try to flesh out some documentation on this sort of thing later.

station384:
I'm very interested in this pull.    I was considering something along these lines while looking at the exchanges,  the thought struck me,  why can't a synthetic exchange pull from a real exchanges data, The more I thought about it, the more I came to the conclusion that it would take a number of mods to trade to support it, which I wasn't ready to tackle.  then I notice this pull, put a smile on my face :)  

Going to pull it (once I figure out how to do a remote fetch of a pull in git) and check it out. 

My initial concerns are does it duplicate the outputs (JSON files) from the normal sim as this effects a lot of the work I've been doing in genetic_backtester

glennfu:
It shouldn't affect output it any way. Sim still runs the same sim, just with slightly different logic in coordinating the timing of events and slightly different price/fee/size calculations.

lib/engine.js
defkev: Thats why i don't work on six different thinks at a time, well at least not in the same branch.

If you don't intend to do a PR for it simply stage the (for us) obsolete lines and include it once you are ready.

Git will take care of putting everything together again.

glennfu: I'm not sure why you're saying it's obsolete. It works correctly as is. It's just a command line parameter. There's no code missing.

defkev: If you say its ready why not simply include it in the PR then?

Update README and add it to trade/sim module.exports option so people actually know the option exists and what it does.

Will i usually discourage mixed commits its still better than adding something nobody knows about.

glennfu: That makes sense. I've added the documentation to those locations. I'll try to be better about splitting up work in the future. It was especially tricky this time.

defkev (approved):

lib/engine.js
defkev: :+1:

glennfu:
Just squashed and rebased to solve merge conflicts

DeviaVir (approved):
:+1: fantastic

krystophv:
I've read it through and didn't see anything glaring that would need to be remedied. I'm a little swamped with work at the moment so I can't go through it super thoroughly, but if DeviaVir and defkev didn't find anything else wrong, then I'm unlikely to without firing up a debugger and stepping through all the code. I certainly like how much this cleans up engine. I think it's probably good to get it going in `unstable` to iron out any kinks that show up in real use. 👍

PR #1383, opened 2018-02-20
move all tablib indicators to a promise and replace calculation usage in strategies as this function does not handle async behavior

Haehnchen:
we have several problem in talib implemention. json decode / encode are too slow. also `calculate` does not provide a callback and is not asyn safe. therefore in `onPeriod` soemtimes the talib result is not include because its faster then the `calculate` calling

### DIFF
--- a/extensions/strategies/ta_ema/strategy.js
+++ b/extensions/strategies/ta_ema/strategy.js
@@ -19,7 +19,6 @@ module.exports = {
   },
 
   calculate: function (s) {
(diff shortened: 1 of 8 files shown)


DeviaVir (approved):
:+1: interesting, nice catch

archaistvolts:
There is a problem with ta_ema/strategy.js line 52 using `Promise.all([]).then(result => {})`.  `result` is an array of results from all the promises, and should be treated as such.  I just used `.then(([result]) => {})` (dereference?) since there is only one result at this point.

PR #1385, opened 2018-02-20
CEXIO: Fix buy maker size calculation

### LINKED ISSUE #923: cexio.btc-usd Place order error: Insufficient funds
------------------------
### System information
- **OS Platform and Distribution :  Linux Ubuntu 16.04
- **Zenbot version : 4.0.5
- **NodeJS version : v8.9.3
- **Selectors : cexio.BTC-USD

### Describe the problem
Error: Place order error: Insufficient funds.
Just for Buy in live trading, and manually b or B, Sell Work !

defkev:
Closes already closed #923

CEXIO (at least) always expects the taker fee will submitting a buy order which causes a `ìnsufficient funds` error if buy_pct (+ makerFee) is greater than or equal to 100% - takerFee

To workaround this "limitation" we will use the taker fee for the buy order size calculation if `buy_pct = 100`

Note that this has no impact on the actual fee paid if the order gets filled, its just to get the order accepted.

This is an exchange specific workaround, so if you suspect that other exchanges have the same "problem" try adding `makerBuy100Workaround: true` to the exchange object in exchange.js

### DIFF
--- a/lib/engine.js
+++ b/lib/engine.js
@@ -471,7 +471,11 @@ module.exports = function (s, conf) {
                 let to_buy_pct = n(so.buy_pct).subtract(held_pct).value()
                 buy_pct = to_buy_pct
               }
-              fee = so.order_type === 'maker' ? s.exchange.makerFee : s.exchange.takerFee
+              if (so.order_type === 'taker' || buy_pct === 100 && s.exchange.makerBuy100Workaround) {
+                fee = s.exchange.takerFee
+              } else {
+                fee = s.exchange.makerFee
+              }
               trade_balance = n(s.balance.currency).divide(100).multiply(buy_pct)
               tradeable_balance = n(s.balance.currency).divide(100 + fee).multiply(buy_pct)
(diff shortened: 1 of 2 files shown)


lib/engine.js
```diff
@@ -471,7 +471,11 @@ module.exports = function (s, conf) {
                 let to_buy_pct = n(so.buy_pct).subtract(held_pct).value()
                 buy_pct = to_buy_pct
               }
-              fee = so.order_type === 'maker' ? s.exchange.makerFee : s.exchange.takerFee
+              if (so.order_type === 'taker' || buy_pct + s.exchange.takerFee <= 100 && s.exchange.makerBuy100Workaround) {
```
DeviaVir: I think this makes `taker` the default, vs `maker` what it used to be. Not sure if it matters.

DeviaVir (approved):

lib/engine.js
defkev: You are right and i'd rather not change it, dunno if its relevant though.

DeviaVir (approved):
:+1:

PR #1398, opened 2018-02-21
Fixes #1380 - binance formatting for assets

### LINKED ISSUE #1380: increment is improperly implemented on Binance
Take EOS/BTC. EOS is an integer coin. `increment` is thus specified as `"1.00000000"` as pulled from `stepSize` from the Binance output.

However, we have engine calculations like:
`n(quote.bid).subtract(n(quote.bid).multiply(s.options.markdown_buy_pct / 100)).format(s.product.increment, Math.floor)`

Now this is taking the EOS stepSize, and passing it to a format of BTC. Luckily, for `numbro`, a format of `"1.00000000"` is functionally equivalent to `"0.00001000". It only cares about how many digits are represented. This means that the current implementation of `increment` is entirely meaningless.

Later on, when you have size calculations, you see things like this: `size = n(s.balance.currency).multiply(buy_pct).divide(100).subtract(size).divide(price).format('0.00000000')`

This gives you values like `9.14567431` when it should be `9` since EOS is an integer coin. Luckily Binance corrects this and spits back `9` after you place an order, but it still screws up calculations within Zenbot and leads to other errors, like trying to buy more than you can afford if what Zenbot wants to buy is very close but not quite an integer.

It's simple to convert `stepSize` to how `numbro` needs it to be like this:
```
    for (i = stepSize.length - 1; i > 0; i--) {
      if (stepSize[i] === '0') stepSize = stepSize.slice(0, i)
    }
```

However, this breaks lots more things because like I mentioned `increment` is passed as the format to BTC, instead of EOS, so the wrong thing gets formatted. I could just reverse this, but I don't understand the original intent so I'd risk breaking more than I fix.

Anyone have any insights on this?

glennfu:
Original issue: #1380, I don't know if this is Binance specific, but it's all I'm using so it's all that's been "fixed". When trading EOS, an integer coin, Zenbot was still fully formatting it as an 8 precision decimal for all buys and sells, which is impossible. Live mode handles this decently because `getBalance` gets the right stuff back and it knows when you buy/sell to truncate at the exchange level, but it messes up sim calculations quite a bit. Hopefully this fixes it.

### DIFF
--- a/lib/engine.js
+++ b/lib/engine.js
@@ -397,11 +397,10 @@ module.exports = function (s, conf) {
             trade_balance = n(s.balance.currency).divide(100).multiply(buy_pct)
             tradeable_balance = n(s.balance.currency).divide(100 + fee).multiply(buy_pct)
             expected_fee = n(trade_balance).subtract(tradeable_balance).format('0.00000000', Math.ceil) // round up as the exchange will too
-            console.log(`s.product.asset_increment: ${s.product.asset_increment}`)
             if (buy_pct + fee < 100) {
-              size = n(tradeable_balance).divide(price).format(s.product.asset_increment ? s.product.asset_increment : '0.00000000', Math.floor)
+              size = n(tradeable_balance).divide(price).format(s.product.asset_increment ? s.product.asset_increment : '0.00000000')
             } else {
-              size = n(trade_balance).subtract(expected_fee).divide(price).format(s.product.asset_increment ? s.product.asset_increment : '0.00000000', Math.floor)
+              size = n(trade_balance).subtract(expected_fee).divide(price).format(s.product.asset_increment ? s.product.asset_increment : '0.00000000')
             }
             msg('preparing buy order over ' + fa(size) + ' of ' + fc(tradeable_balance) + ' (' + buy_pct + '%) tradeable balance with a expected fee of ' + fc(expected_fee) + ' (' + fee + '%)')
           }
@@ -451,7 +450,7 @@ module.exports = function (s, conf) {
         else if (signal === 'sell') {
           price = nextSellForQuote(s, quote)
           if (!size) {
-            size = n(s.balance.asset).multiply(so.sell_pct / 100).format(s.product.asset_increment ? s.product.asset_increment : '0.00000000', Math.floor)
+            size = n(s.balance.asset).multiply(so.sell_pct / 100).format(s.product.asset_increment ? s.product.asset_increment : '0.00000000')
           }
           if ((s.product.min_size && Number(size) >= Number(s.product.min_size)) || (s.product.min_total && n(size).multiply(price).value() >= Number(s.product.min_total))) {
             if (s.product.max_size && Number(size) > Number(s.product.max_size)) {
@@ -822,7 +821,7 @@ module.exports = function (s, conf) {
 
             s.api_order = api_order
             if (api_order.filled_size) {
-              order.remaining_size = n(order.size).subtract(api_order.filled_size).format(s.product.asset_increment ? s.product.asset_increment : '0.00000000', Math.floor)
+              order.remaining_size = n(order.size).subtract(api_order.filled_size).format(s.product.asset_increment ? s.product.asset_increment : '0.00000000')
             }
           }
           syncBalance(function () {
--- a/lib/engine.js
+++ b/lib/engine.js
@@ -397,10 +397,11 @@ module.exports = function (s, conf) {
             trade_balance = n(s.balance.currency).divide(100).multiply(buy_pct)
             tradeable_balance = n(s.balance.currency).divide(100 + fee).multiply(buy_pct)
             expected_fee = n(trade_balance).subtract(tradeable_balance).format('0.00000000', Math.ceil) // round up as the exchange will too
+            console.log(`s.product.asset_increment: ${s.product.asset_increment}`)
             if (buy_pct + fee < 100) {
-              size = n(tradeable_balance).divide(price).format('0.00000000')
+              size = n(tradeable_balance).divide(price).format(s.product.asset_increment ? s.product.asset_increment : '0.00000000', Math.floor)
             } else {
-              size = n(trade_balance).subtract(expected_fee).divide(price).format('0.00000000')
+              size = n(trade_balance).subtract(expected_fee).divide(price).format(s.product.asset_increment ? s.product.asset_increment : '0.00000000', Math.floor)
             }
             msg('preparing buy order over ' + fa(size) + ' of ' + fc(tradeable_balance) + ' (' + buy_pct + '%) tradeable balance with a expected fee of ' + fc(expected_fee) + ' (' + fee + '%)')
           }
@@ -450,7 +451,7 @@ module.exports = function (s, conf) {
         else if (signal === 'sell') {
           price = nextSellForQuote(s, quote)
           if (!size) {
-            size = n(s.balance.asset).multiply(so.sell_pct / 100).format('0.00000000')
+            size = n(s.balance.asset).multiply(so.sell_pct / 100).format(s.product.asset_increment ? s.product.asset_increment : '0.00000000', Math.floor)
           }
           if ((s.product.min_size && Number(size) >= Number(s.product.min_size)) || (s.product.min_total && n(size).multiply(price).value() >= Number(s.product.min_total))) {
             if (s.product.max_size && Number(size) > Number(s.product.max_size)) {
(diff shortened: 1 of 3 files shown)


glennfu:
In another take, trading 10 EOS at a time, the fees kill you pretty hard at the start. Because you end up with a "dust" EOS after your first trade, it's basically as if you only ever had 9 because the first one goes to what amounts to an up-front deposit of fees. Using BNB in this scenario for fees is drastically more appealing, but implementing a BNB fee structure into Zenbot is a lot more complicated than I know how to deal with. In the meantime, this PR makes things more accurate if you're not using BNB.

lib/engine.js
```diff
@@ -397,10 +397,11 @@ module.exports = function (s, conf) {
             trade_balance = n(s.balance.currency).divide(100).multiply(buy_pct)
             tradeable_balance = n(s.balance.currency).divide(100 + fee).multiply(buy_pct)
             expected_fee = n(trade_balance).subtract(tradeable_balance).format('0.00000000', Math.ceil) // round up as the exchange will too
+            console.log(`s.product.asset_increment: ${s.product.asset_increment}`)
             if (buy_pct + fee < 100) {
-              size = n(tradeable_balance).divide(price).format('0.00000000')
+              size = n(tradeable_balance).divide(price).format(s.product.asset_increment ? s.product.asset_increment : '0.00000000', Math.floor)
             } else {
-              size = n(trade_balance).subtract(expected_fee).divide(price).format('0.00000000')
+              size = n(trade_balance).subtract(expected_fee).divide(price).format(s.product.asset_increment ? s.product.asset_increment : '0.00000000', Math.floor)
```
station384: Not sure how this is working.   at first glance it looks like 
if s.product.asset_increment is true it returns s.product.asset_increment else it returns '0.00000000'
are just checking if there is a value?  there always should be an increment.

Math.floor returns a function which is odd to me. will format execute the function using the value from the previous functional link?

Am I missing something here?  does n have its own format and its not utilizing the base string format or am I just ignorant of something.

lib/engine.js
```diff
@@ -397,10 +397,11 @@ module.exports = function (s, conf) {
             trade_balance = n(s.balance.currency).divide(100).multiply(buy_pct)
             tradeable_balance = n(s.balance.currency).divide(100 + fee).multiply(buy_pct)
             expected_fee = n(trade_balance).subtract(tradeable_balance).format('0.00000000', Math.ceil) // round up as the exchange will too
+            console.log(`s.product.asset_increment: ${s.product.asset_increment}`)
             if (buy_pct + fee < 100) {
-              size = n(tradeable_balance).divide(price).format('0.00000000')
+              size = n(tradeable_balance).divide(price).format(s.product.asset_increment ? s.product.asset_increment : '0.00000000', Math.floor)
```
station384: I think we should define what asset_increment is.  I can see its usefulness in other exchanges, and we should update those for consistency.
But we do need the definition at least as this property is now taken into account by the engine.

What is the purpose, and some sample values if the exchange doesn't directly return it should it be 0 or the same as increment

lib/engine.js
station384: looking at this a little more your doing a test here because you introduced a new property to the products definition and can't expect to have all exchanges to have that property.  
The Math.floor question I still have though.

lib/engine.js
glennfu: I have no knowledge or experience with the way other exchanges handle this, and I seem to be the only one who has been bitten by this bug before so apparently the previous behavior hasn't bothered anyone. That's why this conditional preserves the previous behavior for all other scenarios.

station384: Your not the only one.   I ran into the symptoms you described while on bittrex trading a coin that was I believe 100 minimum,  I couldn't get it to ever work.  In fact that indecent got me digging into zenbot's code.

lib/engine.js
station384: Answered my question about the Math.
from the numbro docs: 
You can even round the result with your own rounding function (default is Math.round).

numbro(1000.234).formatCurrency("", Math.ceil);
//$1001

numbro(1000.234).formatCurrency("", Math.floor);
//$1000

station384 (approved):
This looks good and could solve issues on other exchanges with these conditions.   

We do need to come up with a definition for the products object (*pleeeese someone else do it :)*) that is in the wiki.

lib/engine.js
```diff
@@ -397,10 +397,11 @@ module.exports = function (s, conf) {
             trade_balance = n(s.balance.currency).divide(100).multiply(buy_pct)
             tradeable_balance = n(s.balance.currency).divide(100 + fee).multiply(buy_pct)
             expected_fee = n(trade_balance).subtract(tradeable_balance).format('0.00000000', Math.ceil) // round up as the exchange will too
+            console.log(`s.product.asset_increment: ${s.product.asset_increment}`)
```
defkev: Use our pseudo msg() instead of console.log()

lib/engine.js
```diff
@@ -821,7 +822,7 @@ module.exports = function (s, conf) {
 
             s.api_order = api_order
             if (api_order.filled_size) {
-              order.remaining_size = n(order.size).subtract(api_order.filled_size).format('0.00000000')
+              order.remaining_size = n(order.size).subtract(api_order.filled_size).format(s.product.asset_increment ? s.product.asset_increment : '0.00000000', Math.floor)
```
defkev: See below.

lib/engine.js
```diff
@@ -450,7 +451,7 @@ module.exports = function (s, conf) {
         else if (signal === 'sell') {
           price = nextSellForQuote(s, quote)
           if (!size) {
-            size = n(s.balance.asset).multiply(so.sell_pct / 100).format('0.00000000')
+            size = n(s.balance.asset).multiply(so.sell_pct / 100).format(s.product.asset_increment ? s.product.asset_increment : '0.00000000', Math.floor)
```
defkev: See below.

lib/engine.js
defkev: See below.

lib/engine.js
defkev: Fee really is the only thing which should get rounded differently, leave the math to the bot and numbro as this messes with our previous calculation.

If the fix is "exchange specific" consider implementing it like e.g. `s.exchange.makerBuy100Workaround` instead of changing the default behavior.

defkev (changes requested):

lib/engine.js
glennfu: It shouldn't be exchange specific. It should be "coin that isn't 8 precision" specific, which lots of exchanges have. I just don't know how that's handled on all other exchanges so I've done this to create a starting point for it.

lib/engine.js
glennfu: I'll just delete that line

glennfu:
@defkev I can't seem to respond to your last comment inline where you say `Fee really is the only thing which should get rounded differently`. The entire reason for this PR is to fix the fact that `size` is computed incorrectly, and the 4 lines you highlighted are all computing `size`. If I take those out, then my bot will be broken still. In live, it places orders for impossible size values, and in sim, the calculations are busted because it's assuming that impossible size values are actually accepted.

lib/engine.js
defkev: So its exchange-coin-specific?

In that case do the Math.floor only if s.product.asset_increment exists, as we expect all coins to have a precision of 8 by default.

defkev:
See my above inline response on what i would like to see changed.

I simply don't like the idea fixing stuff by altering the "default" behavior of the bot (which seems to be working for the majority of users) because a single exchange is doing something "differently" than others.

See e.g.  #1385 i could have implemented this as the default (which would work) but that doesn't mean it would've made sense.

glennfu:
Maybe it's not showing up right for me but I'm not understanding what your actual suggestion is. As far as I'm aware, none of my changes will affect exchange but Binance. Am I wrong about that?

defkev:
The `Math.floor` will always get executed, regardless of `s.product.asset_increment` as it always is the second argument of numbro's `format()`

glennfu:
Going to the numbro docs it looks like `Math.round` is the default. Do you think that's more correct in this instance?

glennfu:
Also even with `Math.floor` in place, it's still rounding to 8 decimals by default.

station384:
I think the idea is, it shouldn't round,  floor or ceiling if s.product.asset_increment does not exist skip the possibility of altering it.  at least that is what I am getting out of the message chain.

defkev:
If your change works with round remove floor and leave it as the default.

Flooring the result partially annihilates what i fixed with #1376

@station384 :+1:

glennfu:
OK I'll take that part out, but just to be clear, rounding IS happening and has been happening always. numbro's `format` function assumes `Math.round` by default. If I take out the `Math.floor`, it'll just use `Math.round` instead. I'll re-test my code with `Math.round` as default and see how it behaves.

glennfu:
OK as far as I can tell in testing, `Math.floor` and `Math.round` have no difference in practice here, so I've pushed the change.

defkev (approved):
Better

DeviaVir (approved):

PR #1405, opened 2018-02-22
Live vs Sim broke paper mode, hopefully this fixes it

glennfu:
I don't ever use paper mode so I'm not 100% sure what I'm fixing, but it seems ok. In `unstable` right now running in paper mode is basically the same as running live mode, so it'll perform real trades. This fixes that.

### DIFF
--- a/lib/engine.js
+++ b/lib/engine.js
@@ -21,12 +21,15 @@ module.exports = function (s, conf) {
 
   let so = s.options
   if(_.isUndefined(s.exchange)){
-    if (so.mode == 'sim') {
+    if (so.mode !== 'live') {
       s.exchange = require(path.resolve(__dirname, '../extensions/exchanges/sim/exchange'))(conf, s)
     }
     else {
       s.exchange = require(path.resolve(__dirname, `../extensions/exchanges/${so.selector.exchange_id}/exchange`))(conf)
     }
+  }
+  else if (so.mode === 'paper') {
+    s.exchange = require(path.resolve(__dirname, '../extensions/exchanges/sim/exchange'))(conf, s)
   }
   s.product_id = so.selector.product_id
   s.asset = so.selector.asset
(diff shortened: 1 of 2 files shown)


lib/engine.js
```diff
@@ -21,13 +21,16 @@ module.exports = function (s, conf) {
 
   let so = s.options
   if(_.isUndefined(s.exchange)){
-    if (so.mode == 'sim') {
+    if (so.mode !== 'live') {
```
Shawn8901, line 24: this line should also catch line 31. or just to so.mode == 'sim' || so.mode == 'paper'   (so everyone knows whats going on)

glennfu: It doesn't. In paper mode, `s.exchange` is already defined so the check on line 23 is false.

Shawn8901: in paper so.mode isnt paper? (which is also so.mode !== 'live')

Shawn8901: ok, then i dont understand when line 27 is reached. but that's none of your business.

glennfu: Maybe in `mode === 'live'` or `mode === 'train'` which is another mode I've never used but exists.

Shawn8901: yeah but why is the exchange defined in paper in line 23, but not in live when reaching the line? Whats what i meant. But as i said, no need for explanation,

glennfu: I think it actually is defined there in live as well because `trade.js` defines it, and `trade.js` is what's running for both live and paper. So maybe my check is only REALLY catching against train mode? Which, again, I'm not sure in that case that I'm doing this right.

DeviaVir (approved):

lib/engine.js
mpskovvang: This change could be the reason to the #1410 issue.

PR #1412, opened 2018-02-23
Make same rules on binance retry for both buy and sell

glennfu:
My Zenbot got stuck last night pretty much permanently when it missed a sell going through. It kept a stale order alive, in a permanent stuck loop because it had nothing to sell, but never gave up trying. Meanwhile in another thread the bot had long since moved to 'buy' mode and was trying to buy. The Binance exchange has a rule in for buy mode to abort retrying in this event. This rule was absent for sell, but with no clear reason. I've added this to prevent my bot from getting stuck needlessly.

Anyone see any reason they shouldn't have consistent rules for both buy and sell?


DeviaVir (approved):

DeviaVir:
No, and your change makes sense.

PR #1417, opened 2018-02-24
Replace stackoverflow

DeviaVir:
I created this subreddit:
https://www.reddit.com/r/zenbot/

The idea is that all issues tagged with "questions" will be closed and redirected to this subreddit. There are now so many issues that it is no longer a useful tool. This change is designed to alleviate that.

Thoughts?


defkev:
I personally don't really mind wherever people use stack or reddit for questions/discussions as long as they DO use it and keep the Issues for, well, actual issues.

On top, any issue not using the template should be closed as invalid, period.

IMHO the bare minimum, to even get an idea what people are doing:

1. What branch? Stable or Unstable, if unstable what commit?
2. Any changes made to conf-sample.js, not the whole thing but only what differs from the defaults.
3. The command line used, including everything, not just selector and exchange
4. The complete error message, if any

Furthermore, consider closing inactive issues for the unstable branch after a grace period, some of the stuff still open probably doesn't even apply to the current codebase anymore.

phatwila:
@DeviaVir 
Cool, I just made a post to kick things off.

PR #1419, opened 2018-02-24
wait till trade queue is drained before exiting engine

### LINKED ISSUE #1416: Strategy simulation in onPeriod do not take callback into account since latest refactoring
All talib strategies and indicators need a callback as calculations are async. therefore we moved all the calculation from `calculate` to `onPeriod` as this provides a callback. But onPeriod is not using the callback in simulations in the right way since the latests refactorings in engine code. 

If we run a simulation 14 days it stops already before. you can test this behavior with the following code. if you higher sleep time simulator stops more earlier. it looks like some code snippet is not waiting for the callback 

```javascript
    function sleep (time) {
      return new Promise((resolve) => setTimeout(resolve, time));
    }

onPeriod: function (s, cb) {
    Promise.all([sleep(500)]).then(function() {
      console.log("aaa")
      cb()
    })
}
```

krystophv:
Currently the engine exit will allow exiting the process while there are still trades in the processing queue. This will check the queue and, if there are trades still in there (as is usually the case for strategies that require async onPeriod execution), then it will hook into the queue's `drain` callback which will execute after all trades have been processed.

Should resolve #1416 

Also a couple undefined variables and tiny formatting issues that it wouldn't let me commit till I fixed.

### DIFF
--- a/commands/sim.js
+++ b/commands/sim.js
@@ -268,6 +268,7 @@ module.exports = function (program, conf) {
               return getNext()
             }
             engine.exit(exitSim)
+            return
           } else {
             if (reversing) {
               cursor = lastTrade.orig_time
--- a/commands/trade.js
+++ b/commands/trade.js
@@ -549,7 +549,9 @@ module.exports = function (program, conf) {
             }
             if (botStartTime && botStartTime - moment() < 0 ) {
               // Not sure if I should just handle exit code directly or thru printTrade.  Decided on printTrade being if code is added there for clean exits this can just take advantage of it.
-              printTrade(true)
+              engine.exit(() => {
+                printTrade(true)
+              })
             }
             session.updated = new Date().getTime()
             session.balance = s.balance
--- a/lib/engine.js
+++ b/lib/engine.js
@@ -374,6 +374,7 @@ module.exports = function (s, conf) {
         msg('error getting balance')
       }
       getQuote(function (err, quote) {
+        let reorder_pct, fee, trade_balance, tradeable_balance, expected_fee
         if (err) {
           err.desc = 'could not execute ' + signal + ': error fetching quote'
           return cb(err)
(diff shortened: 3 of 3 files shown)


station384:
this should also apply to the --runFor parameter so event it the run length times out,  it doesn't exit till all trades are completed.

edit:  I really SHOULD proof read what I write before I submit..   At times I read it later, like right now  20hrs later and I'm like.  what? --runfor and "so event it"?   MAKE SENSE MAN!

DeviaVir (approved):

krystophv:
@station384 I don't see that parameter anywhere in the code. I thought I remembered someone having it setup for darwin and, if that's the case, this would be fine there. The sim won't exit and, thus, darwin won't process it, until the queue is finished.

station384:
it is in the trade command described in line 55


if option has a value it sets botStartTime the evaluation for the exiting is on line 550.   it calls printTrade(True)  if true is passed,  trade will exit.

krystophv:
Ok. I wrapped that exit in the `engine.exit` function so it behaves the same as `sim` in that regard and will flush the queue before exiting.

DeviaVir (approved):

glennfu:
@krystophv Now that you have `let fee` here:
https://github.com/DeviaVir/zenbot/blob/unstable/lib/engine.js#L381

fee is then undefined on line 388, breaking #1406 

Now when running live my bot does:
```
buy order placed at 0.00140810 BTC
...
buy order timed out, adjusting price
price changed, resizing order, NaN% remain
```
Because fee is `undefined` on line 388.

krystophv:
@glennfu Where was `fee` defined earlier? How could it be anything except undefined at 388?

glennfu:
The same way that size is defined, by a previous loop in the code... I think!

krystophv:
The 

> I think!

is precisely why poorly defined variable scope is a very bad idea. All these variables had no formal definition and perhaps you're correct and they are being automatically hoisted in the execution VM. So it needs to be defined above the function scope? 

`size` is defined as a function parameter

glennfu:
I honestly have no idea how it worked before, I just know it doesn't work today.

glennfu:
Workaround I'm testing now. At line 388 change:
```
if (signal === 'buy') {
    reorder_pct = n(size).multiply(s.buy_order.price).divide(s.balance.currency).multiply(100).add(fee)
} else {
```
To:
```
if (signal === 'buy') {
    fee = so.order_type === 'maker' ? s.exchange.makerFee : s.exchange.takerFee
    reorder_pct = n(size).multiply(s.buy_order.price).divide(s.balance.currency).multiply(100).add(fee)
} else {
```

krystophv:
Seems like https://github.com/DeviaVir/zenbot/pull/1443 might be a solution.

Issue #1397, opened 2018-02-21
Bollinger trade


werxes:
Looks like strategy Bollinger does NOT trading, at least there no output while manual/paper. it is ok while SIM is ruining.

When parameter manual/paper, looks like the no watch on market, no any output console each on period(1m, 5m, etc).

werxes:
this.option('min_periods', 'min. number of history periods', Number, 52)

this line is missing in Bollinger strategy.

Shawn8901:
I cant see where this strategy uses the parameter. if i am wrong, please give the file + line number.

i agree that its one of the essential parameters of zenbot, but why should the strategy claim a usage of it, if it does not? 
 With min_periods you p.e. control the size of your preroll dataset. so if bolliger size is on default (which is 20) and you dont have a min_period of 20 you will actually need 20 live periods before bollinger can be calculated. 
If you have set it to 20 the first live period will be the first which can be calculated.
But thats not a problem of the strategy itself, its more a missing "default" parameter of zenbot, so a misconfiguration in that case.
if another strategy also uses min_periods (for some reason) you have a proper default value, and everything works.

DeviaVir:
Dear issue reporter,

We have slightly changed our github issue policy and would now kindly request folks that have questions that they ask them in our zenbot subreddit.

You can find it here: https://reddit.com/r/zenbot

This issue will be closed, but if you disagree with your ticket being marked as a question feel free to leave a comment defending your case.

Thanks for contributing time and effort!

Greetings,
🤖

PR #1425, opened 2018-02-25
Add min_periods to bollinger

brucetus:


station384:
If I'm correct,  this doesn't need to be added to the strategy,   this is an option that is used in the trade command and the sim command.   all options in those commands apply to all strategies.

Shawn8901:
See #1397 and our discord Discussion. I have the same opinion as station384. A better resolution would be a default value in trade.js/sim.js or in the config file

PR #1428, opened 2018-02-27
GeneticBacktester Floating StartDate

station384:
add: --floatScanWindow
When used the time windows used for scanning, the data will change with
each generation allowing each successive generation to be within the
timeframe of —days selected.

Example with —days 1 —floatScanWindow  the start time frame will be
2018-1-1 00:01 after one day the start time will be 2018-1-2 00:01

without —floatScanWindow
the start will remain at 2018-1-1 00:01

add: --ignoreLaunchFitness
previous fitness is ignore and launch files are always updated to the
current best generation as long as the fitness is > 0.

This helps when using —floatScanWindow as over time the best fitness
yesterday may be in reality the worst fitness today.  so the launch
files need to update with the best one, be it greater or less than the
previous, as long as it is profitable.


DeviaVir (approved):

LoneWolf345:
I would like to discuss this.  I feel it might be better to simply ignore fitness results older than the --days parameter.  Have you already considered this and whats your opinion?

station384:
No I didn’t consider that, but that is a good idea. I’ll try and implement in it once I get back in my state (currently out of state on a job)

LoneWolf345:
are you back yet :P

PR #1429, opened 2018-02-27
getCursor can be different per exchange

krystophv:
In `--paper` mode, the current `sim` exchange is broken for `GDAX` due to the fact that, for whatever reason, the GDAX exchange was implemented using the trade ID as the cursor instead of the time, thus when the exchange looks internally for new trades, it's comparing the wrong number to determine if the trades are new or not. This will point the `sim` exchange's `getCursor` function directly to the real exchange's function. 

Additionally, attempts to simplify matters a little bit by not creating a configured exchange instance in specific command files that also create an `engine` instance, as `engine` will create an exchange instance.

(and again adding the variable definitions that were preventing me from committing `engine` because of eslint errors... not sure how people can commit files with eslint problems still - probably need to add linting checks to Travis)

### DIFF
--- a/commands/balance.js
+++ b/commands/balance.js
@@ -17,7 +17,6 @@ module.exports = function (program, conf) {
     .action(function (selector, cmd) {
       var s = {options: minimist(process.argv)}
       s.selector = objectifySelector(selector || conf.selector)
-      s.exchange = require(`../extensions/exchanges/${s.selector.exchange_id}/exchange`)(conf)
       s.product_id = s.selector.product_id
       s.asset = s.selector.asset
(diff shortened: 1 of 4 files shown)


DeviaVir (approved):

mpskovvang:
Should a similar change be made to getBalance? Paper trading gets the balance of SIM exchange and not the real exchange.

krystophv:
I don't believe so. We're not looking for the _real_ balance in paper trading. The `sim` exchange handles all the "trades" and maintains the balance internally.

mpskovvang:
You're right, that makes a good sense.

PR #1432, opened 2018-02-27
Partial fills

glennfu:
Currently the sim has no way of processing partial orders. This mean you can simulate a low volume market that moves at like 1 BTC/min, buying with 700 BTC and show amazing and impossible gains. With this change, you can actually sim with different balance amounts and see how well you would do, and make sure your strategy properly handles partial orders the same way it would need to in live.

The second commit here makes sure `size` in the engine is always recomputed. I believe this was a bug prior to this, but after adding partial fills, and thus leaving open orders out there longer, the problem was exacerbated. Basically, you put out a buy, compute size based on `buy_pct=100`, then as the engine ticks, and price moves up, but size isn't being recomputed, you easily now have a `price * size` scenario that's higher than your actual balance. The order goes through, and the sim tells you you have a negative balance. Taking out the `if (!size)` checks, the problem is resolved, as sim is recalculated every time a new order adjustment is considered.

The only hangup on this PR is really the `win/loss` calculation. Oddities are:

1. You can place 1 Buy order, and let's say it takes 3 trades to fill, and the prices moves down 3 times for each fill. Then place 1 Sell order, and it completes below the first 2 buy fills but above the last Buy. It will count as 1 win, 0 losses, even if the sell is lower than your first 2 buys, just so long as it is higher than your last buy.

2. You place 1 Buy order, and it goes through quickly. Then you place 1 Sell order. It partially fills 90% of your order at a huge profit, sweet! Then the price tanks and your bot goes into a panic and sells the remaining 10% of your order at a very slight loss. This will track as 0 wins, 1 loss, even though you actually profited overall in this exchange.

Part of me says this is all ok. If you have a sell lower than your buy, it's a loss, and if it's higher it's a win. The profit is still calculated correctly overall. Error rate is just then a bit more... complicated.


defkev (approved):
> size in the engine is always recomputed. I believe this was a bug prior to this

Prior to #1406 reorder used a fixed size.
Will i am not 100% certain this should now indeed be obsolete.

glennfu:
I'm going to back and try put put the `if (!size)` stuff back in and test with your re-order code and see if it solves the problem I was having.

glennfu:
Definitely have something still wrong now. Going to investigate further:
screen shot 2018-02-27 at 11 57 40 am

defkev:
I suppose thats sim-only as we could never end up with a negative balance in live trading, the exchange won't allow us to.

If so something is off with your balance calculation as a order could never get sized over 100% unless you explicitly configure the bot to do so, which we currently don't check.

glennfu:
Found it: `price = n(price).add(n(price).multiply(so.avg_slippage_pct / 100)).format('0.00000000')`

I'm trading NANO with `0.01000000 BTC`. Buy order gets placed at `0.00127950 BTC` for size of 7.81 NANO. This computes as `0.009992895 BTC`. Slippage calculation pushes that over my `0.01 BTC`.

So I need a better way to handle slippage!

glennfu:
Note, as I'm not the one who implemented slippage, and I don't really understand it, I have no idea what to do for this situation.

defkev:
Slippage is simply the amount of price movement (up/down depending on buy/sell) we tolerate during a order.

Positive slippage = loss = bad
Negative slippage = win = good (rarely happens but possible)

e.g. using your data, placing a initial **buy** order at 0.00127950 BTC (the original price), then the price moves up to 0,001292295 BTC (+1% - this usually takes a couple of reorders) so we adjust the price (cancel and reorder) and if your max_slippage is set to (or below) 1% the reorder would trigger slippage protection and cancel the signal, so not place the reorder

Now since there is no "price movement" in sim-mode (imho this could be simulated by simple Math.random()) we use a fixed slippage of 0.045% on every order.

glennfu:
That explanation of what slippage IS makes sense based on what I thought, but I still don't know what to do with it differently. If the effect of slippage needs to be responded to with the ability to cancel, does that mean this needs to be implemented somewhere in the engine?

defkev:
> does that mean this needs to be implemented somewhere in the engine

Its already in engine.js around the code you touched in this commit.

For sim its a different story as there is no "real" slippage, protection will never kick in, well not unless you set so.avg_slippage_pct = max_slippage of course.

glennfu:
I know that `so.max_slippage_pct` is in the engine, but I'm asking what to do with `so.avg_slippage_pct`. Should we just remove it?

defkev:
avg_slippage_pct is the average slippage to apply in sim-mode, so it's handled in sim.js as we don't want/need it in live/paper, where we have a "real" slippage.

glennfu:
`avg_slippage_pct` is loaded as a value in sim.js, but it's not used at all in sim.js. It's ONLY used in the line that I mentioned, which is the problem I'm asking about. The code handles it improperly in sim mode, leading to negative balances. That's what I'm trying to solve.

defkev:
Looks like @krystophv did something about this in #1375

defkev:
Furthermore you have avg_slippage in sim/exchange during price calculation.

glennfu:
Yes that's literally what I copied/pasted from in order to highlight the problem here in this conversation. I'll sync up with @krystophv to see if he can help on this.

glennfu:
So @krystophv's changes from #1375 are actually already merged into this from over a week ago, so it doesn't address this problem.

I spoke with him about some ideas but we don't really have a good path forward to solve that for now. In the meantime while that's brainstormed separately, I think this PR is good to go forward. It doesn't cause or solve the balance going negative bug, so that'll just simply have to be addressed at another time.

defkev:
Mind filling a new issue and (maybe) have @DeviaVir add a "ToDo" or "Confirmed" label so we don't loss track of stuff like this.

glennfu:
I'll do that now!

DeviaVir (approved):
LGTM. Are we happy to merge this?

PR #1433, opened 2018-02-27
Revamp Auto Backtester

glennfu:
I wanted to use the Auto Backtester, and saw it's super outdated. I abstracted the recent changes I did to Darwin and shared them with the Auto Backtester. In order to accommodate that restructuring, I've moved the definition of `phenotype` to the individual strategies. Now Darwin and the Auto Backtester can both pull from the same definition of value ranges.

**As a result, anyone with custom strategies will need to implement the `phenotype` function in order to enable backtesting for their strategy.**

I put notes at the top of the file to explain, but in general, you do very much like you'd do to convert a sim command to a darwin command, giving you:

`./scripts/auto_backtester/backtester.js --selector=binance.ABC-ZYX --rest=1 --of=2 --args=3`

Then you add `step_size` to your liking, such as 10, and then remove 1 or 2 parameters that you want unlocked for backtesting. Assuming those parameters have ranges defined in your strategy's phenotype, it'll step through them in even intervals testing all combinations. In the end you're given a CSV, much like you would with Darwin, with a bit of extra clarity in the columns for the fields you chose.

### DIFF
--- a/extensions/strategies/bollinger/strategy.js
+++ b/extensions/strategies/bollinger/strategy.js
@@ -1,6 +1,7 @@
 var z = require('zero-fill')
   , n = require('numbro')
   , bollinger = require('../../../lib/bollinger')
+  , Phenotypes = require('../../../lib/phenotype')
 
 module.exports = {
   name: 'bollinger',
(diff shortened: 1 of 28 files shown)


DeviaVir (approved):
This change, from the backtester to include it in strategies directly, makes a lot of sense to me. I'll wait for others to chime in as well before merging this.

PR #1437, opened 2018-02-28
fix(package): update jasmine to version 3.1.0

DeviaVir:
Closes #1297


DeviaVir:
same problems, fails randomly at times.

PR #1438, opened 2018-02-28
Pause function

leozzo:
Hi all,

it's possible to add a "pause" from trading, for example after a stop loss?

If there is a huge dump, if i put the stop loss features to ON then the bot probably will buy again after few minute. I would like to tell to the bot "wait a bit, make the dump finishing".

Maybe a pause after a buy. after a sell, and after a stoploss

Thank you


DeviaVir:
This PR does not make sense...

PR #1443, opened 2018-02-28
Store fee in s for use at reorder

defkev:
To help solve #1439 merge this first.

### DIFF
--- a/lib/engine.js
+++ b/lib/engine.js
@@ -343,7 +343,7 @@ module.exports = function (s, conf) {
   // 8. if not filled after timer, repeat process
   // 9. if filled, record order stats
   function executeSignal (signal, _cb, size, is_reorder, is_taker) {
-    let price, buy_pct, sell_pct
+    let price, fee, buy_pct, sell_pct
     delete s[(signal === 'buy' ? 'sell' : 'buy') + '_order']
     s.last_signal = signal
     if (!is_reorder && s[signal + '_order']) {
--- a/lib/engine.js
+++ b/lib/engine.js
@@ -270,6 +270,7 @@ module.exports = function (s, conf) {
       s[type + '_order'] = {
         price: opts.price,
         size: opts.size,
+        fee: opts.fee,
         orig_size: opts.size,
         remaining_size: opts.size,
         orig_price: opts.price,
@@ -280,6 +281,7 @@ module.exports = function (s, conf) {
     let order = s[type + '_order']
     order.price = opts.price
     order.size = opts.size
+    order.fee = opts.fee
 
     order.product_id = s.product_id
     order.post_only = conf.post_only
@@ -385,7 +387,7 @@ module.exports = function (s, conf) {
         }
         if (is_reorder && s[signal + '_order']) {
           if (signal === 'buy') {
-            reorder_pct = n(size).multiply(s.buy_order.price).divide(s.balance.currency).multiply(100).add(fee)
+            reorder_pct = n(size).multiply(s.buy_order.price).divide(s.balance.currency).multiply(100).add(s.buy_order.fee)
           } else {
             reorder_pct = n(size).divide(s.balance.asset).multiply(100)
           }
@@ -526,6 +528,7 @@ module.exports = function (s, conf) {
       placeOrder(signal, {
         size: size,
         price: price,
+        fee: fee || null,
         is_taker: is_taker,
         cancel_after: so.cancel_after || 'day'
       }, function (err, order) {


lib/engine.js
```diff
@@ -385,7 +387,7 @@ module.exports = function (s, conf) {
         }
         if (is_reorder && s[signal + '_order']) {
           if (signal === 'buy') {
-            reorder_pct = n(size).multiply(s.buy_order.price).divide(s.balance.currency).multiply(100).add(fee)
+            reorder_pct = n(size).multiply(s.buy_order.price).divide(s.balance.currency).multiply(100).add(s.buy_order.fee)
```
theswim, line 390: `reorder_pct` becomes `buy_pct` and it used for calculating `tradeable_balance` and `size`.
and `fee` is considering again when calculating `tradeable_balance` and `size`.
so, I thought no need to add `fee` to `reorder_pct`.

and `so.buy_pct` and `reorder_pct` are becomes to `buy_pct` and it used for sizing.
If `so.buy_pct` is not contains `fee` then `reorder_pct` should not contains `fee`.

If I misunderstood could you explain it for me?

defkev: fee during a reorder is whatever fee was will creating the order, so we resize the order properly.

tradeable balance simply is the, well, amount of balance we are actually able to trade.

As no exchange would want to allow us to trade past the point the won't be able to pay our fees (can't pay if none is left after filling the order) we deduct it from balance prior to filling the order BUT only account for it if balance + fee actually excites 100%

DeviaVir (approved):
:+1:

glennfu:
@defkev @DeviaVir I've been testing this today, and I still end up with output like: `'price changed, resizing order 100.3247189% remain'`

It then tries to buy for more than 100% of what I have, and gets an out of funds error. What if we add `buy_pct = Math.min(reorder_pct, so.buy_pct)` after that?

Or is there a better solution? To be honest I don't understand the who reorder_pct concept at all. The bot knows the available balance, the on hold balance, why don't we just set `size = null` and let it recalculate it every time? I don't think I understand the reasoning behind remembering the previously used percentage.

defkev:
If you simply let the bot recalculate the size on every signal we would always use `buy|sell_pct` - even on a reorder which already got partially filled.

Take a snapshot of `s.buy|sell_order` and post it here to get a idea what numbers we are talking about.

glennfu:
One second look while I'm seeing the > 100%'s, it seems to be correct for the instances I can find. It might have been a red herring so I might have been causing the out of funds errors with something else. If I can find more clear evidence again I'll come back to this. In the meantime I think it might be working fine.

glennfu:
OK I found it. I've forced `s.exchange.makerFee` to be 0 for simplicity:
```
Starting balance: 0.00000000 XRB  0.01039813 BTC

2018-03-01 09:14:04 - buy order timed out, adjusting price
2018-03-01 09:14:04 - remaining size: 6.95

s.buy_order: {"price":"0.0014997","size":"6.93","fee":null,"orig_size":"6.95","remaining_size":"6.95","orig_price":"0.0014959","order_type":"maker","cancel_after":"day","product_id":"XRB-BTC","post_only":true,"order_id":1459,"orig_time":null,"time":null,"local_time":1519913623674,"status":"open"}

2018-03-01 09:14:04 - price changed, resizing order, 100.23836016668382% remain
2018-03-01 09:14:04 - preparing buy order over 6.95000000 XRB of 0.0104229 BTC (100.23836016668382%) tradeable balance with a expected fee of 0.00000000 BTC (0%)

2018-03-01 09:14:04 - placing buy order...
Trying to buy for 0.010410405 vs actual balance of 0.01039813 BTC
2018-03-01 09:14:04 - not enough balance for buy, aborting
```

Note no partial order stuff is happening here. It's an order that starts with `so.buy_pct = 100` and it's just trying to re-adjust the order after the price changes.

defkev:
Whats your BTC balance?

glennfu:
Top line of my log there I specified it: `Starting balance: 0.00000000 XRB  0.01039813 BTC`

defkev:
nvm. :man_facepalming: 

```
"size":"6.93",
...
"orig_size":"6.95",
"remaining_size":"6.95"
```

As size differs from the original size i presume the order already got resized previously.

> 6.95 * 0.0014997 / 0.01039813 * 100 + null = 100,2383601666838

The math is definitely correct, yet it doesn't make sense.

Almost like we are missing to update something in s...._order on a reorder.

Let me spin up my dev box and take a closer look.

defkev:
I suspect something is fishy around

`// order timed out and needs adjusting`

with size vs remaining_size.

Crunching your data it boils down to:

1. Create order over 6.95 XRB at 0.0014959 BTC using 99.98% of 0.01039813 BTC balance
2. Resize order to 6.93 XRB at 0.0014997 BTC using 99.95% of 0.01039813 BTC balance
3. Resize order again to 6.95 XRB :thinking: 

As the size moved up (6.93 -> 6.95) i would assume the price moved down (instead of up) which would explain why we overshoot the second reorder by 0.23%

Which brings me back to my initial assumption that
> we are missing to update something in s...._order on a reorder

which is simply order.remaining_size.

This way we end up doing:

1. Create order
2. Resize order (based on price)
3. Resize order back to remaining_size as size (from 2) and remaining_size differ

Let me prep a PR

defkev:
Done.

#1451

Do a Test-Drive and let me know in the PR how it goes.

PR #1456, opened 2018-03-02
Add Lolex for better simming of timing

glennfu:
Recently I added an `eventQueue` to replace `setTimeout` because I thought it would give us better control of timing. However, while that works for the engine, it doesn't abstract well to other parts of the app that need asynchronous timing, like the sim exchange. I found the `lolex` gem which can take over native time calls, and now we can bring back `setTimeout` and have fully fast-forwardable asynchronous timing for our sims. Hooray!

I also implemented a 100ms latency to all sim exchange calls. After adding this, my sim vs live comparison have gotten MUCH closer together. Now when I debug one side having a missing or extra trade from the other, I can account for it by realizing that the order adjust happened some number of seconds later or earlier. This is of course impossible to make equivalent since network speeds are never consistent or predictable, but I think this is as close as the sim will ever get to matching live. I'm really happy with the state of sim vs live now.

### DIFF
--- a/extensions/exchanges/sim/exchange.js
+++ b/extensions/exchanges/sim/exchange.js
@@ -2,20 +2,34 @@ let path = require('path')
   , n = require('numbro')
   , _ = require('lodash')
 
-
 module.exports = function sim (conf, s) {
 
+  let latency = 100 // In milliseconds, enough to be realistic without being disruptive
   let so = s.options
   let exchange_id = so.selector.exchange_id
   let real_exchange = require(path.resolve(__dirname, `../${exchange_id}/exchange`))(conf)
 
   var now
-  var balance = { asset: so.asset_capital, currency: so.currency_capital }
+  var balance = { asset: so.asset_capital, currency: so.currency_capital, asset_hold: 0, currency_hold: 0 }
 
   var last_order_id = 1001
   var orders = {}
   var openOrders = {}
-  var last_trade
+  let debug = false // debug output specific to the sim exchange
+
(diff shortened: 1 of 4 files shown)


DeviaVir:
Code looks good, I'll run some tests before merging.

Thanks for all your work going into making the sim better, so far!

defkev:
Just out of curiosity, and without diving into any of this, but why not simply make use of Promises if the sim/exchange relies on asynchronous execution?

glennfu:
Mostly this was an attempt to solve the difference between sim and live without rocking the boat too much. We get to keep the structure that is already in place, while also giving us the ability to "speed up time" so that we can sim in a way that matches the structure of live.

As for a new structure to change the behavior of live, perhaps using Promises, I'm very open to any of that. In fact, I've got some major concerns. For example, with Binance, you can sometimes get an error back suggesting an order wasn't placed, and yet it was placed. This can lead to Zenbot not tracking/adjusting the order, and later being surprised when the order was filled. A solution would involve less `getOrder` calls and more calls to check the entire open order list. Right now if you have a callback loop on an order, you don't have any real hooks into its various `setTimeout` "chains". If you wanted to grab an order in the middle of its various behaviors, you can't. You just have to hope that the next function that gets triggered will bail out or something.

All that being a long winded way to say I wish we could get rid of the whole `setTimeout` thing altogether and have behaviors that are a bit more tightly controlled and intentional in each moment. However, that's too much of a job for right now!

DeviaVir:
Ran a few tests, looks good. Thanks for contributing again!

PR #1465, opened 2018-03-05
Ensure results with negative vsBuyHold don't get backward fitness

glennfu:
If you have 2 Darwin results that perform worse than `vsBuyHold`, previous fitness calculation will weight them backwards, meaning the one that makes less money will have the higher fitness. This happens because vsBuyHold is a negative number. Being a percentage, simply adding 100 ensures it will never be negative but still preserves the ordering of results, and solves the reverse order problem of results that aren't as good as `vsBuyHold`. Thanks to @krystophv for leading me to this.

### DIFF
--- a/scripts/genetic_backtester/phenotype.js
+++ b/scripts/genetic_backtester/phenotype.js
@@ -45,8 +45,8 @@ module.exports = {
         let items = ['simple', 'low', 'trend']
         let index = Math.floor(Math.random() * items.length)
         r[k] = items[index]
-      } 
-      
+      }
+
     }
     return r
   },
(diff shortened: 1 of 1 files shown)


krystophv:
To expand on how/why this works: when taking the win/loss ratio into account, the fitness calculation will multiply the `vsBuyHold` by a scaling factor between 0.5 and 1.0 depending on how the win/loss ratio is balanced. When `vsBuyHold` is positive, but has many losses vs wins, it will effectively have 1/2 the fitness of an identical `vsBuyHold` that has many wins vs losses. If `vsBuyHold` is negative, then multiplying by 1/2 actually increases the value (as it gets closer to zero), while multiplying by 1 would not move the value. This has the effect of rewarding a very poor win/loss ratio for negative `vsBuyHold` values. Since `vsBuyHold` is bounded by -100 (in the case that the sim lost _all_ the money), we can simply linearly scale the whole set by 100 and bound it by 0. Since it's a linear scaling, we keep the correct order.

DeviaVir (approved):

Issue #1463, opened 2018-03-04
First in last out for buy and sell order


stinkyjames:
### System information
- **Have I written custom code (as opposed to using zenbot vanilla)**: I can understand and modify basic codes, like change the output from "vs. buy hold" to "bot vs. buy hold"
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: 
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.3 LTS
Release:	16.04
Codename:	xenial

- **Zenbot version** (commit ref, or version): current master
- **Zenbot branch**: master
- **NodeJS version**: v8.9.4
- **Python version (when using a python script)**: not using python script
- **Exact command to reproduce (include everything)**: NA
- **Did I make any changes to conf-sample.js?**: I changed the conf.js to my desired settings

### Describe the problem
I have made changes to the engine.js file to enable it to make multiple buys without waiting for a sell to happen by reducing "held_pct" to 0.

When the sell side is triggered, it only looks at the last buy price instead of performing a first-in last-out of the buy prices. I'm guessing a counter needs to be added to the buy price for the sell price to reference, such as buy_price_1, and buy_price_2, and buy_price_3.. or buy_price_x where x is incremented for each buy and when it triggers a sell it will look at the most recent buy_price_x , for example buy_price_3, and makes sure to reference buy_price_3, and then nulls buy_price_3 and moves on to the buy_price_2 if another sell is triggered to sell by referencing buy price_2. It correctly works with the last buy price but I would like it to work for all the previous buys until only the first buy is left to be referenced as the last buy price.

basically the scenario will play out as follows:

first buy trigger:
buy_price_1

sell trigger
sell references buy_price_1

buy trigger:
buy_price_1

buy trigger:
buy_price_1 exist
create buy_price_2

sell trigger
sell references buy_price_2
nulled/deletes buy_price_2

buy trigger:
buy_price_1 exist
create buy_price_2

buy trigger:
buy_price_1 exist
buy_price_2 exist
create buy_price_3

sell trigger
sell references buy_price_3
nulled/deletes buy_price_3

sell trigger
sell references buy_price_2
nulled/deletes buy_price_2

buy trigger:
buy_price_1 exist
create buy_price_2

sell trigger
sell references buy_price_2
nulled/deletes buy_price_2

sell trigger
sell references buy_price_1

sell trigger
sell references buy_price_1

That's in. First-in Last-out for buy_price when a sell triggered.

[...]

PR #1471, opened 2018-03-05
Always use lowest/highest price for buy/sell loss protection

### LINKED ISSUE #1389: Buy will not execute if last successful trade was also a buy, sells work fine
### System information
- **Have I written custom code (as opposed to using zenbot vanilla)**: Nope, not at all.
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: WSL Ubuntu 16.04
- **Zenbot version**: 4.1.0
- **NodeJS version**:  8.something or 9 both have problem
- **Python version**: 2.7 or 3+ Both have problem
- **Exact command to reproduce**: Any live trading scenario, with default .conf 
./zenbot trade gdax.LTC-USD --buy_pct 50 --sell_pct 50
will produce a perfect example of the issue.

### Describe the problem
Zenbot will trade perfectly as long as it follows the buy...sell...buy pattern, however, I would like it to buy if there is a price improvement and it will not do this, for example sell...buy...buy...sell. It will however, sell...sell...buy.
I would like to use 50% percent of the currency balance to buy the asset, and I do not have a problem with it buying ever smaller 50% chunks of the balance, this is the desired behavior. However, it buys with 50% and will not execute another buy until a sell has succeeded. Multiple sell works just fine, and it will sell down until the minimum has been reached. Buys will also buy until the balance remaining is too small, but only if it keeps selling between buys. I also can not trigger a manual buy if the last executed trade was a buy, no errors occur in debug and it seems the request is simply ignored.

Short of it, I would like to know why the bot will not perform 2 or more buys in a row if the price improves and triggers another buy signal.

The sim and backtester, both work as though multiple buys occurred in a row, without a sell to separate them.

This occurs with any strategy, or USD-asset pair on gdax, though it honestly seems as though it will occur with any live trade scenario.

### Source code / logs
./zenbot trade gdax.LTC-USD --buy_pct 50 --sell_pct 50
Debug provides no information or errors. The bot will not buy if the last signal that succedded in trading was a buy. It will also not allow a manual buy in the same scenario, the request is ignored silently. Multiple sells can succeed in a row, and a manual sell trade can be executed after a successful sell.

defkev:
On consecutive buys/sells the bot would always use the last price for loss protection, which could lead to a potential (sic) loss if the last price is below/above (depending on the order type) any previous price of the same order type.

To overcome this we will lookup all previous orders (of the same type) and adjust last price to the highest/lowest within the previous orders.

Will this isn't a true FIFO (as we don't account for any amount bought/sold) it goes in the same direction by always "aiming" for a profit.

This completely removes held_pct, which i presume was supposed to accomplish the same but seems to be causing a bug preventing the bot from doing consequtive buys.

Also some linter fixes will i am already at it.

Tester welcome!

Closes #1389 & #1463

### DIFF
--- a/commands/trade.js
+++ b/commands/trade.js
@@ -409,7 +409,7 @@ module.exports = function (program, conf) {
               my_trades.find({selector: so.selector.normalized, time : {$gte : trades[0].time}}).limit(0).toArray(function (err, my_prev_trades) {
                 if (err) throw err
                 if (my_prev_trades.length) {
-                  s.my_prev_trades = my_prev_trades.slice(0).sort(function(a,b){return a.time + a.execution_time > b.time + b.execution_time ? -1 : 1}) // simple copy, most recent executed first
+                  s.my_prev_trades = my_prev_trades.slice(0).sort(function(a,b){return a.time + a.execution_time < b.time + b.execution_time ? -1 : 1}) // simple copy, less recent executed first
                 }
               })
             }
--- a/lib/engine.js
+++ b/lib/engine.js
@@ -433,8 +433,8 @@ module.exports = function (s, conf) {
               size = s.product.max_size
             }
             msg('preparing buy order over ' + fa(size) + ' of ' + fc(tradeable_balance) + ' (' + buy_pct + '%) tradeable balance with a expected fee of ' + fc(expected_fee) + ' (' + fee + '%)')
-            let lastest_low_sell = _.chain(trades).dropRightWhile(['type','buy']).takeRightWhile(['type','sell']).sortBy(['price']).head().value() // return lowest price
-            let buy_loss = lastest_low_sell ? (lastest_low_sell.price - Number(price)) / lastest_low_sell.price * -100 : null
+            let latest_low_sell = _.chain(trades).dropRightWhile(['type','buy']).takeRightWhile(['type','sell']).sortBy(['price']).head().value() // return lowest price
+            let buy_loss = latest_low_sell ? (latest_low_sell.price - Number(price)) / latest_low_sell.price * -100 : null
             if (so.max_buy_loss_pct != null && buy_loss > so.max_buy_loss_pct) {
               let err = new Error('\nloss protection')
               err.desc = 'refusing to buy at ' + fc(price) + ', buy loss of ' + pct(buy_loss / 100)
@@ -481,8 +481,8 @@ module.exports = function (s, conf) {
             if (s.product.max_size && Number(size) > Number(s.product.max_size)) {
               size = s.product.max_size
             }
-            let lastest_high_buy = _.chain(trades).dropRightWhile(['type','sell']).takeRightWhile(['type','buy']).sortBy(['price']).reverse().head().value() // return highest price
-            let sell_loss = lastest_high_buy ? (Number(price) - lastest_high_buy.price) / lastest_high_buy.price * -100 : null
+            let latest_high_buy = _.chain(trades).dropRightWhile(['type','sell']).takeRightWhile(['type','buy']).sortBy(['price']).reverse().head().value() // return highest price
+            let sell_loss = latest_high_buy ? (Number(price) - latest_high_buy.price) / latest_high_buy.price * -100 : null
             if (so.max_sell_loss_pct != null && sell_loss > so.max_sell_loss_pct) {
               let err = new Error('\nloss protection')
               err.desc = 'refusing to sell at ' + fc(price) + ', sell loss of ' + pct(sell_loss / 100)
--- a/lib/engine.js
+++ b/lib/engine.js
@@ -209,7 +209,7 @@ module.exports = function (s, conf) {
       if (s.my_trades.length) {
         last_trade = s.my_trades[s.my_trades.length - 1]
       } else {
-        last_trade = s.my_prev_trades[0]
+        last_trade = s.my_prev_trades[s.my_prev_trades.length - 1]
       }
       s.last_trade_worth = last_trade.type === 'buy' ? (s.period.close - last_trade.price) / last_trade.price : (last_trade.price - s.period.close) / last_trade.price
       if (!s.acted_on_stop) {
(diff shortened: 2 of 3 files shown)


stinkyjames:
Is it possible to log buy_price and coin_amount for each buy action? Then when the sell side is executed it references the buy_price with the associated coin_amount and will only sell above the buy_price with the same coin_amount? So we can have a few buys and when the sell side happens, it will sell each coin_amount at a higher value of the same coin_amount.

For example:

a buy, buy, sell, sell happens

buy = a.b_p @ $100 for 0.001 BTC
buy = b.b_p @ $200 for 0.0025 BTC

sell = a.s_p @ $210 for 0.0025 BTC
sell = b.s_p @ $125 for 0.001 BTC

I want the sell side to only sell the exact amount of coins purchased at the specified markup in the conf.js file. This would correctly address our FILO needs.

defkev:
As i wrote, this indeed isn't a true FIFO as we don't account for any amount bought/sold.

The PR doesn't alter the bots trading behavior, it will simply prevent the bot from potential loss by always buying/selling with a profit by aiming for the highest/lowest price within all previous opposite orders up to the last identical signal.

e.g. on a `buy@0.2 - buy@0.1 - sell@0.2 (instead of last) - sell@0.3 - buy@0.2 (instead of last)` the bot will use the highest buy price of the first two buys during the first sell and the lowest sell price during the third buy.

Doing a proper FIFO/LIFO is definitely doable, as we keep a list of executed orders it would simply be a matter of using the completed order size closest to the current market price will creating the order on a signal to use for buy|sell_pct

But as a change like this would alter the bots default trading behavior it would need to be implemented as a opt-in feature, something like a `accounting = 'FIFO' // or LIFO`

stinkyjames:
How much work would be required to do a FILO for the sell side to reference the buy side? I think with this option, we would be able to make money with all our trades, even when the price of the coins drop, the bot can still make purchases as it goes down, increasing its opportunities to make money as the coins value return at a later time, instead of waiting for the coin's value to return to its pike price of the first buy in when the coin value started to drop.

I've also tested the engine.js modification from above and I wasn't able to get it to work. The buy and sell shows complete as soon as it is placed but it appears the bids are just cancelled.

defkev:
Now this sounds more like a strategy you are talking about.

> The buy and sell shows complete as soon as it is placed but it appears the bids are just cancelled

Thats not possible to be related to this commit, it simply alters the last_price which only affect loss protection.
I am running this on four different bots for ~1 day and haven't seen anything odd yet after multiple trades.

stinkyjames:
Let me look at my strategy.js. I’m using trend_ema and corrected the “buy” and “sell” trigger.

In the meantime, do you plan to do a true FILO option? Also, thank you for being so responsive.

defkev:
I am always interested in stuff increasing possible profit :grin: 

But as i am not an Accountant i would need to do some research first to make sure i don't mess it up in the progress.

stinkyjames:
We will always make money if we treat every buy seperate from one another. As value of the coin dips and bot continues to pick up coins, there will be more opportunities for it to sell the coins back and make profit as it increases instead of getting stuck at the top of the dip, it will pick up more coins on the way down and sell those coins as the value recovers to above the purchase price of each buy. If you draw a graph or put it in a spreadsheet you will see we can make more on the dips than if the coin only raises in value. If the coin is bullish, it’s better to hold, but when the coin is bearish it’s good to make purchases on the way down. Lock the price at the highest price limits how much and how often we can sell. If we have it sell above each of the buy, we can gain profit on each buy and its respective amount.

defkev:
In normal "Trade" mode the bot only trades if he receives a signal from the strategy used, such a signal can only get emitted on a period (the length of a period is defined by `--period|period_length`)

The size of an order can be defined by `--buy|sell_pct`

Assuming you set both to 10% (without defining `--buy_max_amt`) the bot would sell 10% chunks of your asset on each period and buy using 10% chunks of your balance, again on each period, depending on the strategy.

Usually you would want to sell on a down trend/bearish and buy on up trend/bullish to maximize your profit.

stinkyjames:
You're correct about buying on trends. In a buy buy sell sell scenario, what I'm trying to prevent is the second sell from selling lower than the first buy. One solution is with FILO, or have it remember the purchase price and amount of the coin, and only sell that amount for higher than the purchase price for every buy. It will work for buy buy buy sell sell sell scenarios since it will treat every buy independent of the other buys. So if one buy was purchased at a high price, it will hold it until the price returns to above that level while still being able to buy more coins and selling those coins at the lower prices. Basically the bot can still make trades without worrying about it selling the one that was purchased at the high price or wait for the price to return to the high price while buying and selling at the new lower price. If the bot references only the purchases at the highest price, and the price doesn't return to that level, the bot won't make any sell, but if the bot is able to reference each buy price, and only need to beat the last buy price, it will make money for the last purchase, and then move on to the next buy price. If we can truly get a FiLO of the buy prices, it will be so amazing!

defkev:
> what I'm trying to prevent is the second sell from selling lower than the first buy

So do i.

I refactored the whole thing using lodash (should've done this initially :see_no_evil:) to also account for previous buys on a second (+ consecutive) sell and vice versa.
Furthermore it will now include previous trades in the calculation, if available.

This still doesn't cover the amount bought/sold, as this goes in the direction of buy|sell_pct rather than max_buy|sell_loss_pct, which this PR is about.

I'll test drive it for a bit and see how it goes.

Unfortunately my primary exchange (with my test wallet) is currently down for maintenance, so its going to take a while.

stinkyjames:
Awesome! I’m a patient person. As long as you can figure out the first in last out for the sell side to reference the buy side’s first in last out, we will be golden.

stinkyjames:
Every time it cancels, it logs the cancel as a trade. You can see the cancels in the output html in the stat folder. There's something incorrect in the trade.js file. Everything else looks promising. Thank you for the amazing work.

defkev:
> Every time it cancels, it logs the cancel as a trade.

Sounds like #1482

stinkyjames:
Followed #1482 and the issue was resolved. Thank you for the link!

DeviaVir:
@defkev these changes LGTM and don't introduce any issues afaict, ready to merge?

defkev:
Go ahead

DeviaVir (approved):

Issue #1467, opened 2018-03-05
Per Currency Conf with custom DB entry


carpionetwork:
### System information
- **Have I written custom code (as opposed to using zenbot vanilla)**: False
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: CentOS 7
- **Zenbot version** (commit ref, or version): unstable
- **Zenbot branch**: unstable
- **NodeJS version**: 8.3
- **Python version (when using a python script)**: n/a
- **Exact command to reproduce (include everything)**: Detailed below
- **Did I make any changes to conf-sample.js?**: False

### Describe the problem
I have created conf files for each of the three coins (BTC, ETH and LTC) (btc_conf.js, eth_conf.js and ltc_conf.js).  Within each I changed the mongodb entry from zenbot4 to (For BTC (btcbot), For ETH (ethbot) and For LTC (ltcbot)).

Is it better to keep one mongodb zenbot4 that houses all three of the coins or is creating a per coin database fine? 

The reason I am asking this is because I am seeing a random issue where the buy or sell signal will print but neither continues to bought or sold.  Even though at that moment a buy or sell would have been good.  I have sufficient funds for each coin as well as USD for either transaction.

Thank you!

kareha:
Which exchange?
If Bitfinex: Have you ever traded the currency before?
I had the same behavior. See here: https://github.com/DeviaVir/zenbot/issues/695#issuecomment-355825828 and fixed it on my own (works now).

It doesn't matter if you split up the db collections by asset, you can keep all trades in one.

carpionetwork:
Gdax ... have not tried any other exchange.  The bot has traded successfully before but recently it is stuck at printing the buy / sell signals.

As a test I switched to the bollinger strategy to see if there is a buy/sell signal issue with just the trend_ema strategy.  Sidenote: I am using the latest unstable code.

Thanks for the link.  

I'll let you know how it goes with the new strategy after it has had time to settle in for a few hours.

defkev:
> buy or sell signal will print but neither continues to bought or sold

This simply means the bot didn't execute the signal, enable debugging (`L` key) to see why.

The order is `buy - buying - bought` and `sell - selling - sold`

Haehnchen:
[...]

Issue #1423, opened 2018-02-25
Some trades are missed on exchange and completed in bot


kongres:
### System information
- **Have I written custom code (as opposed to using zenbot vanilla)**: strategy: trend_ema + `console.log` lines in some places
- **OS Platform**: Windows Server 2008 R2
- **Zenbot version**: unstable branch https://github.com/DeviaVir/zenbot/commit/ab655e3783a05f8f44d88369c39739c046df7dc7
- **NodeJS version**: 8.9.4
- **Python version**: 2.7
- **Exact command to reproduce**: 
node zenbot.js trade --conf=configs/conf-ETH.js
node zenbot.js trade --conf=configs/conf-XRP.js

Hi!
Firstly, Thank you for the great bot! 
I'm working with this bot and I've noticed one strange behavior. Some trades are not appearing on Poloniex. 
The following screenshot shows this issue:
![image](https://user-images.githubusercontent.com/7360716/36644708-69b3a038-1a6f-11e8-9ea9-a8c89a4156fa.png)

I didn't check if it placed any order during this time, but in the end no any record about this activity in history on Poloniex. It happens not always. Most trades are fine and have records in history on Poloniex. 

I'm not sure but maybe it is related with multi-instances? I've created configs files for each instance. And run the bot in two different instances with two different configs.

kongres:
Happened again:
![image](https://user-images.githubusercontent.com/7360716/36644917-29165b8a-1a72-11e8-992d-7d1ca92879d9.png)

Primea1:
I've had this issue on binance with ETH/USDT. It said it completed a trade, but it never did and just kept saying "sold" over and over. then the bot stopped working completely.

kongres:
I've added logging on `trade` method in `exchange.js` and nothing. It doesn't call the method when show these messages...
Checking other places...

station384:
I've gotten this too.   while working out a bug in hitbtc and I think I may know a way to reproduce the condition that this happens.

This occurred during adjustment of the price of an order, and how the exchange logs orders. 
It didn't happen all the time but enough that I could narrow it down and work around it.
This is HitBTC specific,  but other exchanges may work and report the same ways, or similar.
[...]

PR #1475, opened 2018-03-06
gdax exchange code format and provide some more debug information

Haehnchen:
to allow some more debugging output i have added some logging and did some code formatting on gdax exchange implementation. related to #1467 #1423


DeviaVir (approved):
neat :+1:

Issue #1426, opened 2018-02-26
Hull Moving Average Indicator


hpmmatuska:
Hello,

I use Hull Moving Average Indicator a lot. Probably it may help others as well and I thing, it will be benefit to have such indicator in the list. 

https://www.fidelity.com/learning-center/trading-investing/technical-analysis/technical-indicator-guide/hull-moving-average

I am sorry, but I am not able to made it bymyslef.

Haehnchen:
Simple to implement lets see. we directly not have HMA but its calucate via WMA that is already include via talib
```
HMA= WMA(2*WMA(n/2) − WMA(n)),sqrt(n))
```

PR #1480, opened 2018-03-08
add "Hull Moving Average Indicator" based on tulind library #1426

Haehnchen:
this adds Hull Moving Average Indicator / hma as a strategy #1426 and also tulind as an alternative for talib. which is faster and provides better node api (imho)


defkev:
:+1:  for adding tulind

DeviaVir (approved):
:+1:

also +1 for axing TA

PR #1482, opened 2018-03-09
fix canceled orders in gdax are detected as a successful order in websocket connection and bot does mark it as bought

Haehnchen:
i did same live debug on gdax. it looks like cancel order are handle as successful order. i simply placed an order and canceled it. zenbot jumped into bought with a success message. ;) as the order can also canceled by gdax eg timeout this describes same issue like #1423

_Note: Also there is retry for non websocket connections inside "client.getOrder" but this code can not work anymore "done_reason = 'canceled'" is not used by engine. also other exchange implementations using this conditions but never catched_


DeviaVir (approved):
:+1: great catch

PR #1490, opened 2018-03-12
Hello DeviaVir - Neural Strat enhancement.

ghost:
Just a few changes:
Comparing old prediction vs. new prediction.
Using stock decay, stock depth (2, for a up/down decision possibly increase to create a large pattern-matching trend pattern recognition) per Stanford's website and some stackexchange documentation.

I hope you all are doing well, I threw out my neck a long while back so I've been trying to take it easy.

~~At any rate, this strat still needs tuning in live mode to work with if (!s.in_preroll) but I havent figured this out.~~ <- I think this is resolved. However, it requires time on first run to calculate. But I didn't see any bugs in sim. I'll run it again to check.

However, this thing would be great threaded but I'm sure that forex.analytics and cxxt modules are more profitable.

### DIFF
--- a/extensions/strategies/neural/strategy.js
+++ b/extensions/strategies/neural/strategy.js
@@ -14,18 +14,19 @@ module.exports = {
   name: 'neural',
   description: 'Use neural learning to predict future price. Buy = mean(last 3 real prices) < mean(current & last prediction)',
   getOptions: function () {
-    this.option('period', 'period length - make sure to lower your poll trades time to lower than this value. Same as --period_length', String, '1h')
-    this.option('period_length', 'period length - make sure to lower your poll trades time to lower than this value. Same as --period', String, '1h')
+    this.option('period', 'period length - make sure to lower your poll trades time to lower than this value. Same as --period_length', String, '30m')
+    this.option('period_length', 'period length - make sure to lower your poll trades time to lower than this value. Same as --period', String, '30m')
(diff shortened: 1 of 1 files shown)


LoneWolf345:
Dont forget to update the phenotype at the bottom as well.

ghost:
Could use some help setting decimal type phenotypes, this option is new since I last saw zenbot.

If anyone is familiar please chime in.






Also this thing was insane with a trendline run across 20 BTC pairs on poloniex. I could see 20% in a day (im exaggeurating probably) but this is when I hurt my neck so I stopped development/didn't debug. I just had 1-hour period and like 90-days of backfill data loaded and trained into the neural network and the trendline set to 10/1000?

The signaling I long since forgot though.  I think it was only buying on neural prediction increase and uptrend using the trendline function (just one trendline, no fancy stuff.) And way too many neurons. The downside was I didn't have a "already bought" type variable or function in the program. I've tried to code this in but had no luck.

DeviaVir (approved):
:+1: looks great, let's see it in the wild

DeviaVir:
@juanzolo Interesting.  I merge your changes in so it gets more people testing and more feedback, I think that's the fastest way for improvements to land.

ghost:
Thanks :) To be honest, some of the settings in this strategy are way out of my league. It's too slow for me to want to backtest haha!

PR #1491, opened 2018-03-12
Fix: sim/exchange dynamic fees exception

### LINKED ISSUE #1483: Error on simulation with Cex.io
### System information
- **Have I written custom code (as opposed to using zenbot vanilla)**:
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: ubuntu 16.04
- **Zenbot version** (commit ref, or version): 109411b46ced8d364faa8867e693775fc86a7c8d
- **Zenbot branch**: unstable
- **NodeJS version**: 8.10.0
- **Python version (when using a python script)**:
- **Exact command to reproduce (include everything)**: ./zenbot.sh sim
- **Did I make any changes to conf-sample.js?**: on cex.io

### Describe the problem
Hi, I got this error when I try to do a simulation with Cex.io 

### Source code / Error logs
/home/phil/zenbot/lib/engine.js:61
    s.exchange.setFees({asset: s.asset, currency: s.currency})
               ^

TypeError: s.exchange.setFees is not a function
    at module.exports (/home/phil/zenbot/lib/engine.js:61:16)
    at Command.<anonymous> (/home/phil/zenbot/commands/sim.js:95:20)
    at Command.listener (/home/phil/zenbot/node_modules/commander/index.js:315:8)
    at emitTwo (events.js:126:13)
    at Command.emit (events.js:214:7)
    at Command.parseArgs (/home/phil/zenbot/node_modules/commander/index.js:651:12)
    at Command.parse (/home/phil/zenbot/node_modules/commander/index.js:474:21)
    at /home/phil/zenbot/zenbot.js:46:13
    at FSReqWrap.oncomplete (fs.js:135:15)

defkev:
setFees (if supported by the exchange) instead of erroring out

Closes #1483

### DIFF
--- a/extensions/exchanges/sim/exchange.js
+++ b/extensions/exchanges/sim/exchange.js
@@ -161,6 +161,12 @@ module.exports = function sim (conf, s) {
       }, latency)
     },
 
+    setFees: function(opts) {
+      if (so.mode === 'paper' && real_exchange.dynamicFees) {
+        return real_exchange.setFees(opts)
+      }
+    },
+
     getCursor: real_exchange.getCursor,
 
     getTime: function() {
--- a/extensions/exchanges/sim/exchange.js
+++ b/extensions/exchanges/sim/exchange.js
@@ -162,9 +162,7 @@ module.exports = function sim (conf, s) {
     },
 
     setFees: function(opts) {
-      if (so.mode === 'paper') {
-        return real_exchange.setFees(opts)
-      }
+      return real_exchange.setFees(opts)
     },
 
     getCursor: real_exchange.getCursor,
--- a/extensions/exchanges/sim/exchange.js
+++ b/extensions/exchanges/sim/exchange.js
@@ -162,7 +162,7 @@ module.exports = function sim (conf, s) {
     },
 
     setFees: function(opts) {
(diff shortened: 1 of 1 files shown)


extensions/exchanges/sim/exchange.js
```diff
@@ -161,6 +161,12 @@ module.exports = function sim (conf, s) {
       }, latency)
     },
 
+    setFees: function(opts) {
+      if (so.mode === 'paper') {
```
DeviaVir, line 165: Maybe I'm missing something, but why would we only want to do this in paper? (And not sim?)

defkev: Original implementation had it only in live/paper as the feature relies on a (real time) timer we cannot "sim", but then again it wouldn't hurt.

defkev: Actually it does hurt :grin: 

Just did some sim runs with debugging enabled and noticed that setFees will get called on each new trade day causing us to hammer the exchange with requests into oblivion.

Reverted to only work in paper.

DeviaVir (approved):

PR #1494, opened 2018-03-12
Completely rework --buy_max_amt

defkev:
As much as i like the --buy_max_amt feature i was never a fan of how it was implemented, leave alone the naming being confusing at best as it simply doesn't just account for the amount the bot can buy but how much he is actually able to trade.

* Renames --buy_max_amt to --deposit for sake of clarification
* Using --buy_max_amt will still work but print a deprecated warning
* Instead of using --deposit to size buy orders use it to limit the balance available to the bot
* This allows us to use it together with --buy_pct
* --deposit will still account for asset held to calculate the available currency balance
* So if the asset held is covering --deposit the bot will have 0 currency (0% of currency balance) available, if not currency will be filled to the point it compensates for the difference (up to 100% of currency balance)
* If deposit is exceeding asset + currency the bot will also use 100% of currency balance held
* Deposit will be updated on each price change and new period to ensure the bot always only trades the amount as defined by --deposit
* Profit calculation will no longer go havoc ~~as long as currency balance doesn't fall below whats currently deposited into the bot~~
* If enabled this will add two new columns to the output displaying the total currency balance (in green) as well as how much percentage (gray) of currency balance (0 - 100%) is currently used by the bot
* Works in Live, Paper & Sim mode

### DIFF
--- a/commands/trade.js
+++ b/commands/trade.js
@@ -30,7 +30,7 @@ module.exports = function (program, conf) {
     .option('--asset_capital <amount>', 'for paper trading, amount of start capital in asset', Number, conf.asset_capital)
     .option('--avg_slippage_pct <pct>', 'avg. amount of slippage to apply to paper trades', Number, conf.avg_slippage_pct)
     .option('--buy_pct <pct>', 'buy with this % of currency balance', Number, conf.buy_pct)
-    .option('--buy_max_amt <amt>', 'buy with up to this amount of currency balance', Number, conf.buy_max_amt)
+    .option('--deposit <amt>', 'absolute initial capital (in currency) at the bots disposal (previously --buy_max_amt)', Number, conf.deposit)
     .option('--sell_pct <pct>', 'sell with this % of asset balance', Number, conf.sell_pct)
     .option('--markdown_buy_pct <pct>', '% to mark down buy price', Number, conf.markdown_buy_pct)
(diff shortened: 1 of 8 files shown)


defkev:
Just noticed that this doesn't cover the stats returned by `P` and thus not the profit returned by the API. Might as well add a new asset column to the Web GUI (as deposit is stored in s.balance) will i am at it :thinking: 

Leave it open for now..

defkev:
I finally had some time to polish the whole thing and fix the profit calculation for sims.

Note that the balance (last/end) reported by stats in live/paper/sim mode will now use deposit as its base value instead of the absolute balance if `--deposit` is enabled.

commands/trade.js
```diff
@@ -80,6 +80,10 @@ module.exports = function (program, conf) {
       so.debug = cmd.debug
       so.stats = !cmd.disable_stats
       so.mode = so.paper ? 'paper' : 'live'
+      if (so.buy_max_amt) {
+        console.log(('--buy_max_amt is deprecated, use --deposit instead!\n').red)
+        so.deposit = so.buy_max_amt
+      }
```
DeviaVir, line 86: _nice_

DeviaVir (approved):
LGTM, this is a much requested feature and I think this makes it a lot more clear. Thanks!

PR #1499, opened 2018-03-13
Updated to match what gdaxbot is using.

ghost:
Literally, cookie-cutter code from his first strategy release.

```https://github.com/VertigoRay/GDAXBot/blob/master/strategies/stddev.js```

This leads me to some questions though about the end results:
```  "sell_pct": 99,
  "sell_stop_pct": 0,
  "show_options": true,
  "start": 1519689600000,
  "stats": false,
  "strategy": "stddev",
  "trendtrades_1": 5,
  "trendtrades_2": 1000,
  "verbose": false,
  "simresults": {
    "start_capital": 1000,
    "currency": 1034.81333053,
    "profit": 0.03481333053000003, <- Kind of hard to grasp
    "buy_hold": 883.8052209050325,
    "buy_hold_profit": -0.11619477909496755,
    "total_trades": 9,
    "length_days": 18,
    "total_sells": 5,
    "total_losses": 3,
    "vs_buy_hold": 17.086130071774473
  }
}```

When, the second from last columb catches my eye before zenbot exits sim:
```2018-03-12 18:14:59   9135.26 BTC-USDT   -0.25%        3   55.63 ++         sell  0.00001 BTC  1034.71006842 USDT   +3.48%  +17.02```
```+3.48%``` <- and -> ```+17.02``` Seems almost like they should be "reversed" in documentation.

As vs. Buy Hold would be true profit was my assumption.

### DIFF
--- a/extensions/strategies/stddev/strategy.js
+++ b/extensions/strategies/stddev/strategy.js
@@ -13,7 +13,7 @@ module.exports = {
     this.option('period_length', 'period length, set poll trades to 100ms, poll order 1000ms. Same as --period', String, '5m')
     this.option('trendtrades', 'Trades for array 2 to be calculated stddev and mean from', Number, 1000)
     this.option('min_periods', 'min_periods', Number, 1001)
-	this.option('order_adjust_time' 'Order Adjust Time', Number, 999999)
+    this.option('order_adjust_time', 'Order Adjust Time', Number, 999999)
   },
   calculate: function () {
   },
--- a/extensions/strategies/stddev/strategy.js
+++ b/extensions/strategies/stddev/strategy.js
@@ -4,36 +4,49 @@ var z = require('zero-fill')
   , ema = require('../../../lib/ema')
   , Phenotypes = require('../../../lib/phenotype')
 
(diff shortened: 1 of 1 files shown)


ghost:
As a side note, it would be beneficial to require --days in sim for the most accurate results when adjusting --period

Also seeing by this strategies results, It can seem profitable at first hand, but it makes me want to throw out the idea of speed trading in the first place.

IE I'm currently running a 90-day sim, BTC has been on the downtrend. My vs. buy hold is 85% 3 months in. This concerns me that maybe there should also be a backfill minimum. I have had 0 luck with trading on downtrend because of bitcoins short-term volatility. And ESPECIALLY in speed trading this nets losses. (maybe why it's called a future.)

The bubble already burst, per say. Sadly. And my mongodb is majorly lagging 2 months into the 90-day backtest. its lagging so much, that backtesting the amount of data needed to be backtested could be a major issue. I'm using nodejs9. Going to test with numactl --interleave=all mongod (hopefully it can resolve this)

ghost:
However I tested, this code wasn't profitable.

PR #1504, opened 2018-03-14
Added -runGenerations parameter

theDiverDK:
In order to do some testing, i added a '-runGenerations' parameter.

If defined, it will run for that number of generations.

Also it will show like '=== Simulating generation 1 of 9 ===' if you specified 'runGenerations=10'

The reason it shows 'of 9' is due to it starts with generation=0

Did more code cleanup and fixed a bug due to i remove a needed line :(

### DIFF
--- a/docs/scripts/genetic_backtester.md
+++ b/docs/scripts/genetic_backtester.md
@@ -38,7 +38,7 @@ The following parameters are available when executing darwin.js:
 --use_strategies="all | strategy1,strategy2"                                            # With this parameter, you can choose to test all, some (comma separated), or just one of the available strategies defined within darwin.
 --population="150"                                                                      # Optional - Number of simulation per generation
 --population_data="./simulations/backtest_[simtimestamp]"                               # Optional - Resume backtesting on a previously terminated backtesting session.
---runGeneraions										# Optional - Makes it possible to stop after a number of generations
+--runGenerations									# Optional - Makes it possible to stop after a number of generations
 ```
 
 ## Results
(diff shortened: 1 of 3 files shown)


docs/scripts/genetic_backtester.md
```diff
@@ -38,7 +38,7 @@ The following parameters are available when executing darwin.js:
 --use_strategies="all | strategy1,strategy2"                                            # With this parameter, you can choose to test all, some (comma separated), or just one of the available strategies defined within darwin.
 --population="150"                                                                      # Optional - Number of simulation per generation
 --population_data="./simulations/backtest_[simtimestamp]"                               # Optional - Resume backtesting on a previously terminated backtesting session.
-
+--runGeneraions										# Optional - Makes it possible to stop after a number of generations
```
DeviaVir: typo here: `runGeneraions`

DeviaVir (approved):

docs/scripts/genetic_backtester.md
theDiverDK: Thanks for spotting it.

Made a push with it fixed :)

LoneWolf345:
Would you consider adding a push notification when darwin finishes?

theDiverDK:
Hi LoneWolf345

Actually I have not :)

But I would gladly implement it, if not to hard.

What type of push notification are you thinking about?

My idea with this command, was to say, run for 25 generations, then take the best and start trading, and then start over.

I am planning on adding a --runForMinutes

So you can say, Run for 5000 minutes, trade with best start over.

LoneWolf345:
I think you can trigger a notification and it will push to all configured notifiers.  I was thinking something along the lines of "Darwin has finished running _<number of gens>_ generations and found _<strategy name>_ had the best result with an end balance of _<end balance>_.

Or something like that.  could also replace end balance with ROI.

LoneWolf345:
I also like the --runForMinutes flag as well.  I think it would compliment if people have a different preference.

theDiverDK:
Ahh cool, had not looked into the notification part.

I'll look into it pretty soon. And i'll add the --runForMinutes as well :)

LoneWolf345:
maybe check out this from engine.js
```
  function pushMessage(title, message) {
    if (so.mode === 'live' || so.mode === 'paper') {
      notifier.pushMessage(title, message)
    }
  }
```

PR #1510, opened 2018-03-17
SNYK assessment patches.

ghost:
https://snyk.io/test/github/moongem/zenbot

As good as I could get the zenbot with the least amount of vulnerable dependencies.

I havent tested it fully yet, if anyone wants to test with me be my guest :)

Good to have for reference on what needs to be changed. Some of the packages were forked onto my github and those NPM dependencies were updated. I might try and make PRs for them as a alternative route to updating.

Edit:  I mainly wanted to PR to see what needs to be modified, or, if anyone can help test - maybe with genetic algo/all strats? that would be much appreciated.

I don't have a gemini account but. This version is likely unstable, but needless to say atleast more secure.

Maybe we could get a hardened branch going I can PR into or a development branch.

Have a good day all! :-)


ghost:
Seeing if I can get travis CI to build, removed snyk test from the test in package.json

ghost:
The problematic packages:

https://github.com/MoonGem/tulipnode
https://github.com/MoonGem/poloniex.js
https://github.com/MoonGem/gemini-api-node

PRs were opened in all three forks so they will (hopefully) update the npm modules.

Have a blessed day.

Shawn8901:
In general its a great idea, but wouldnt it be better to get the vulns fixed instead of miving bsck to older versions? This repository has greenkeeper enabled, it will auto create a pr for the "old versions". So it will result in dependency ping pong...
Just my 2 cents

ghost:
Not a bad idea. But, I've actually been thinking of holding a vote to remove poloniex completely as it's where 90% of the vulnerability was. I don't want to bash on polo, but I fear they are running some type of well... Ponzi scheme? Exit scam? I have no problems trading on polo, but when the polo price is low on bitcoin they hodl your withdraw and if the price raises they cancel in hope you'll trade more and they make more dividends. 

Just seems really shady is all. When I make withdraws when the price is going low they always hang up awaiting approval and get cancelled when the price comes back up. Sheer market manipulation that is detrimental to the entire cryptocurency ecosystem. I say we just boot the exchange. Sometimes the servers even lag up to a whole two minutes to place an order.

Shawn8901:
instead of dropping poloniex support, we could switch to cctx instead of this (like bitbtc and therock).

DeviaVir:
I'd be okay with opening a `secure` branch, but I think you removed your fork.

PR #1511, opened 2018-03-17
syncBalance can return quote, no need to get it again

krystophv:
Currently, as part of `executeSignal`, we run `syncBalance` and then immediately run `getQuote` when `getQuote` is actually run _inside_ `syncBalance`... So instead of running `getQuote` twice, we can just have `syncBalance` return the quote (and balance for that matter, to start moving these things away from just mutating a psudo-global) and use that quote. 

Involved removing a level of nesting for most of the **enormous** `executeSignal` function, so turning off whitespace with `?w=1` is recommended for readability.


DeviaVir (approved):
:+1:

PR #1519, opened 2018-03-22
Create ichimoku

brucetus:
Basic ichimoku cloud trading strategy. It buys when the price is above the cloud and sells when the price is inside or below the cloud.


extensions/strategies/ichimoku
```diff
@@ -0,0 +1,59 @@
+var z = require('zero-fill'),
```
DeviaVir: I think you forgot to add the `.js` extension to this file

DeviaVir (commented):
Already merged in `highest` and `lowest` (which are dependencies), so if you could only fix the filename here, this one is good to go as well :+1:

brucetus:
Haha stupid mistake on my part. Fixed it

DeviaVir (approved):
:+1:

PR #1520, opened 2018-03-22
Create highest.js

brucetus:
Highest finds the highest price (AKA highest high) in the past # of periods given by key.


DeviaVir (approved):
:+1:

Issue #1517, opened 2018-03-21
[Feature Request] make an update script


LoneWolf345:
### Describe the Feature
A suggestion on Discord by @shr00mie was to create an update script for zenbot.  He suggested makeing a ./zenbot.sh update command.

This would probably really help out a lot of support issues.

janus007:
I'm just so happy with your usage.

archaistvolts:
Here is a windows version if anyone wants to use it.  Basically the same except for the remove node_modules directory command
```bat
git checkout package.json package-lock.json
git pull

rd /q /s node_modules
npm install && npm update && npm dedupe
```

PR #1523, opened 2018-03-22
Update script for #1517

shr00mie:
for #1517 per @LoneWolf345

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 # the below will work assuming no other custom modifications have been made.
-rm -rf node_modules
+rm -rf node_modules/
 git checkout package.json package-lock.json
 git pull
 npm install && npm update && npm dedupe
--- a/update.sh
+++ b/update.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+# the below will work assuming no other custom modifications have been made.
+rm -rf node_modules
+git checkout package.json package-lock.json
+git pull
+rm -rf node_modules/
+npm install && npm update && npm dedupe
--- a/update.sh
+++ b/update.sh
@@ -3,5 +3,4 @@
 rm -rf node_modules
 git checkout package.json package-lock.json
 git pull
-rm -rf node_modules/
 npm install && npm update && npm dedupe


update.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+# the below will work assuming no other custom modifications have been made.
+rm -rf node_modules/
+git checkout package.json package-lock.json
+git pull
+npm install && npm update && npm dedupe
```
DeviaVir, line 6: should we care about updating scripts here, too?
should it be mentioned in the README?

update.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+# the below will work assuming no other custom modifications have been made.
+rm -rf node_modules/
+git checkout package.json package-lock.json
```
Shawn8901, line 4: can you please explain, why to checkout this files first, and then pull the changes afterwards? Shouldnt both files be updated on a pull as well?

update.sh
Shawn8901: Since there are a lot of questions about getting recent updates, it would be a nice step to supply an update script. but you are right, there should be a reference from the README.md file to this script.

@shr00mie on my opinion it would be perfect to supply also a bat file to update, since zenbot also tries to support native windows support. you could use rmdir instead of rm (for the bat file), rest should be the same.

update.sh
DeviaVir: I think this is mainly to make sure there are no local changes to these files that would mess up a `git pull`

shr00mie: @Shawn8901 you can't pull if modifications have been made. npm update & dedupe will modify both.

<3

update.sh
shr00mie: @Shawn8901 *puts on elitist linux hat while submitting this post on a windows7 box*

shr00mie: @DeviaVir it was just a thought while discussing stuff with @LoneWolf345 in regards to ways to minimize people reporting issues. i would concur with placing it in README, but for the fact that it would require noobs to read the document to begin with, which...well. just go in #general.

shr00mie: alternative would be to build it in as a command so you could run ./zenbot update...maybe have it do a checkout for all affected files based on git response and then update...but that's more effort/time than i have at the moment. at least if it's in the base directory with +x already set, it might allow users, especially those not on the dev side, to just really quickly be able to do a full module wipe, pull, and install.

PR #1525, opened 2018-03-23
Add info comment and phenotypes to ichimoku

brucetus:
Added a little disclaimer/ info as to what this strategy does by default. Also @LoneWolf345 added the phenotypes for use in the genetic_backtester script.


DeviaVir (approved):
:+1:

PR #1535, opened 2018-04-04
[Strategy] Ichimoku Score

LoneWolf345:
This is a new variant on Ichimoku that calculates all the indicators, scores them, weights them to normalize the scale, then adds them all up.  
Accompanying this is a readme that also contains a tradingview.com script that should very closely match.
I have also started a helpers lib that contains some trading view formatted functions that were used in this strategy and will be used for a few more I have in the works.

*This is my first pull request ever so I welcome any constructive criticism.  Be kind.


DeviaVir (approved):
:+1:

Issue #1487, opened 2018-03-12
intfactor in phenotype.js formula error


LoneWolf345:
### System information
- **Have I written custom code (as opposed to using zenbot vanilla)**: no
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Linux Ubuntu 16.04
- **Zenbot version** (commit ref, or version): current
- **Zenbot branch**: unstable

### Describe the problem
Currently it is possible for the formula for intfactor in phenotype.js to evaluate to zero even if its outside the designated range.

### Source code / Error logs
Here is the offending code.
`      } else if (v.type === 'intfactor') {
        // possible 0 value by providing min 0
        if (v.min == 0 && Math.random() <= 0.5) r[k] = 0
        else r[k] = Math.round((Math.random() * (v.max - v.min + 1)/v.factor)*v.factor)
      }`
Here is a possible fix by Shawn
`      } else if (v.type === 'intfactor') {
        // possible 0 value by providing min 0
        if (v.min == 0 && Math.random() <= 0.5) r[k] = 0
        else r[k] = Math.round(((Math.random() * (v.max - v.min) + 1) /v.factor)*v.factor)
      }`

Shawn8901:
Why not creating a pr if you already got some code to fix the issue? :-)

LoneWolf345:
I barely function with git as is.  I'm not confident with I wouldn't push my API keys :)

archaistvolts:
Can anyone help me understand why there is a 50% chance of returning 0 if v.min == 0?  This doesn't seem right to me.  It seems like the chance of returning 0 should be equal to (v.max - v.min) / v.factor.  

```javascript
      else if (v.type === 'intfactor') {
        // possible 0 value by providing min 0
        if (v.min == 0 && Math.random() <= 0.5) r[k] = 0
        else r[k] = Math.round(((Math.random() * (v.max - v.min) + 1) / v.factor) * v.factor)
      }
```

archaistvolts:
@DeviaVir @Shawn8901 I am still having problems with RangeFactor returning out of range values with a current pull.  I am testing this function which seems rock solid so far.  Please consider using this one as I believe its a vast improvement over the current one.  

It works with floating points and floating point factors and gets rid of any funny repeating decimals that show up sometimes with the current solution.  Getting this right is very important to darwin as it can help prevent so many other errors that come up such as the stalling issues i was experiencing due to parameter values being off.  

```javascript
[...]

Issue #721, opened 2017-11-16
Placing STOP LOSS orders for trailing stop instead of limit sell orders


mikeh21:
Feature request.

HI just wanted to know if there is a reason why the bot doesnt use the STOP LOSS order function when using trailing stop? this would make it much easier for the bot to sell the top rather than chasing it up with the limit sell orders it is currently using?... can this be easily changed in the exchange scripts?

    --profit_stop_enable_pct <pct>  enable trailing sell stop when reaching this % profit
    --profit_stop_pct <pct>         maintain a trailing stop this % below the high-water mark of profit

mikeh21:
PS Im referring to GDAX exchange I do not have experience of other exchange but expect the same is possible and would also benefit from this feature

nedievas:
STOP orders usually have taker fees. It this OK?

mikeh21:
Hi, I know on GDAX can place Stop Loss W/ Limit order which is no fees and can be placed below market price to prevent loss. could this be implemented in zenbot to use trailing stop loss to lock in profit in trending market?

manuelpaulo:
Same request! Thanks.

skeebuzz:
Unfortunately, the API doesn't currently support stop-limit. You can put in a stop order, but it turns into a market order when triggered.

shri046:
@skeebuzz I am interested in this feature as well and based on the [discussion here](https://github.com/coinbase/gdax-node/issues/164) it looks it is possible to place such an order, it is just not documented. I have not tested this myself.


```
authenticatedClient.sell( {
  type: "limit",
  product_id: "BTC-USD",
  size: "0.005",
  price: "10100",      // the limit price
  stop: "entry",       // currently undocumented
  stop_price: "10000", // currently undocumented
});
```

skeebuzz:
@shri046 Oooowwww... thanks for that! Good to know!!

skeebuzz:
Gdax now officially supports them: https://docs.gdax.com/#place-a-new-order

obinnaokechukwu:
@shri046 That's not quite right. For sell side it should be stop: "loss" not "entry". For buy side, it's "entry".

yang2lalang:
This looks a very interesting feature that should solve the slippage problem in live trading where order execution is a big mess and you see limit orders filled after 11 minutes

stop limit order should help trigger  price points for strategy.js to act on

kilo42L:
[...]

PR #1538, opened 2018-04-07
bump gdax-node version and change phenotypes.create intfactor

archaistvolts:
This PR only modifies a few lines, changing package.json and lib/phenotypes.js. 

gdax-node has been [updated](https://github.com/coinbase/gdax-node/pull/286) around 11 days ago to support stop losses and entries in version 0.7.0.  I'm not sure how to execute one with zenbot yet, but #721  seems to indicate that it will be possible with this change. 

Also, I've added a few lines to fix phenotypes.RangeFactor in the create function as described in #1487.


archaistvolts:
I just noticed that the previous PR is a bump to gdax version too.  Haha oops.  Well atleast this includes the RangeFactor change too so its not useless.

PR #1549, opened 2018-04-13
 Predict whole candles instead of close price only

ratajczak:
I think that working with whole candles should give better close price prediction because of open, low, high and volume can form some patterns which can be taken in to account when predicting close price.
This would also enable more sophisticated strategies based on candle patterns like buy when last period and predicted period are hammers, etc.


DeviaVir (approved):
:+1: probably best for neural, I agree.

PR #1557, opened 2018-04-22
Improved Docker + auth in database

mgsotelo:
- Previous configuration returned a lot of errors when using docker.
- Added comments on docker-compose.
- Removed entrypoint/command from Dockerfile, it caused an error, and it was not worth it.
- Switched to tutum/mongodb instead of official mongodb because it allows to set user&password from docker-compose instead of going to the command line
- Added adminmongo to visualize the db related things
- Added support for authMechanism (db related) and modified conf-sample.js to support that change (previous configuration led to an error while connecting to the database if someone did not use the connectionString but used the parameters listed in conf-sample)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .DS_Store
+.env
 *.tar.gz
 *.log.*
 t.js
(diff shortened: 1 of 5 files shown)


DeviaVir:
I'll include the authMechanism changes in a followup PR, but I don't see the improvements you've made in regards to docker-compose; the adminmongo is already in the file, and the auth/container changes seem unnecessary.

I'll close this for now, thanks for contributing regardless! :+1:

Issue #1514, opened 2018-03-18
Bittrex Problem


DrMabuse-2017:
### System information
- **OS Platform and Distribution (Linux Debian 9.3 amd64):
- **Zenbot version (version 4.1.0):
- **Zenbot branch *latest*:
- **NodeJS version v8.10.0:
- **Python version (when using a python script) *no*:
- **Exact command to reproduce (include everything) *Dashboard*:
- **Did I make any changes to conf-sample.js? **:

Hello guys...
i have a problem wit bittrex

example:
Step 1. 2018-03-18 20:50:41 - placing sell order...
Step 2. 2018-03-18 20:50:41 - sell order placed at 0.00000349 BTC

I have see the sell order in bittrex !

Step 3. 2018-03-18 20:54:24 - 0.00000348 vs our 0.00000349
Step 4. 2018-03-18 20:54:25 - cancel failed, order done, executing

Now the result from Step 4:
sell order completed at 2018-03-18 20:54:25:
14384.85027394 XVG at 0.00000349 BTC
total 0.0502031 BTC
0.0000% slippage (orig. price 0.00000349 BTC)
execution: 4 minutes

I think the Step 4 is wrong.
	- the order is definitely canceled in bittrex !
	- Step 4: order done is not ok 
	- Step 4: executing is not ok 

The sequence and the order of functions (Step 4) is wrong?
	
The order was definitely not executed!
But in zenbot is the order executing.

Take a look in screenshot

who can tell me what is wrong

The command
./zenbot.sh trade bittrex.XVG-BTC --debug --strategy trend_ema --conf=conf.bittrex.js --trend_ema 21 --period_length 1m --period 1m --max_slippage 0.001 --poll_trades 6000 --order_poll_time 3000 --order_adjust_time 4000 --oversold_rsi_periods=1000 --oversold_rsi=1000 --neutral_rate=0.065  --max_sell_loss_pct=0.85 --max_buy_loss_pct=5 --buy_pct=100 --sell_pct=100 --rsi_periods 1100  --profit_stop_enable_pct 1 --profit_stop_percent 0.5 --markup_sell_pct 0.02  --markdown_buy_pct 0.02 --reset-profit

Many Thanks
Mario
![bittrex_bad_failed](https://user-images.githubusercontent.com/33420293/37570693-50746e9c-2af3-11e8-9c77-ede0a783b578.png)

fdk0:
+1

brucetus:
+1

[...]

PR #1563, opened 2018-04-25
Fix Bittrex error

brucetus:
I fixed the order cancel error in issue #1514. I changed the cancelOrder and getOrder functions. 

I also changed 'fistRun' to 'firstRun'


DeviaVir (approved):
:+1: nice!

PR #1583, opened 2018-05-04
Docker and commandline improvements

TechupBusiness:
Hey, I'm using docker excessively and needed some improvements. Would like to share it to everybody (make and reverse proxy with basic auth is very helpful). And the open port of mongodb is a huge security issue with docker, everybody can connect without any password right now on the default port and check/change your database.


Makefile
```diff
@@ -0,0 +1,79 @@
+ARGS = $(filter-out $@,$(MAKECMDGOALS))
+MAKEFLAGS += --silent
+
+list:
+	sh -c "echo; $(MAKE) -p no_targets__ | awk -F':' '/^[a-zA-Z0-9][^\$$#\/\\t=]*:([^=]|$$)/ {split(\$$1,A,/ /);for(i in A)print A[i]}' | grep -v '__\$$' | grep -v 'Makefile'| sort"
```
kenorb, line 5: Suggestion, two: `grep -v '__\$$' | grep -v 'Makefile'` can be combined into one like:

    grep -v -e '__\$$' -e 'Makefile'

.env-sample
```diff
@@ -0,0 +1,5 @@
+BASIC_AUTH_ZEN=ommm:$apr1$1n0zqO2y$A2ueB9zMWZTYTtfckFhql1
+TRAEFIK_HOST_ZEN=myhost.com
```
kenorb, line 2: Could you use `example.com` instead? Same below. I think it's a better candidate.

DeviaVir (approved):
this is great stuff, thanks for digging all of this out.

DeviaVir:
The two comments are good feedback but they can be PR'd later on, I'll go ahead and merge this now.

PR #1611, opened 2018-06-07
Replaced fs.existsSync(Deprecated) with fs.accessSync

cryptocoinserver:
https://iojs.org/api/fs.html#fs_fs_existssync_path

fs.existsSync(path)
Stability: 0 - Deprecated: Use fs.statSync or fs.accessSync instead.


cryptocoinserver:
There may be more uses of fs.existsSync in the code.

cryptocoinserver:
Can't replace it that easy. My fault :D Im working on a solution.

PR #1621, opened 2018-06-15
Add strategies that utilize TA-Lib and Tulip-Lib

station384:
Adds backing libs,  and zenbot extensions strategies

### DIFF
--- a/extensions/strategies/ta_srsi_bollinger/strategy.js
+++ b/extensions/strategies/ta_srsi_bollinger/strategy.js
@@ -12,22 +12,22 @@ module.exports = {
     this.option('period_length', 'period length, same as --period', String, '5m')
     this.option('min_periods', 'min. number of history periods', Number, 200)
     this.option('rsi_periods', 'number of RSI periods', 14)
-    this.option('srsi_periods', 'number of Stochastic RSI periods', 14)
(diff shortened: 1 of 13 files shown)


station384:
There are duplicate strategies using the two libraries because they both bring different features to the table.
ta-lib is slow,  but is more flexible and extensive,  it allows calculations using  EMA,  DEMA, WMA, TRIX, T3, and more when calculating bollinger bands and stoch,  but it is slow.

Tulip is much faster, on some items its 3x faster, but can only calculate using EMA. 

Note:
All these strategies use the Promise pattern, so they only calculate on a candle period, this causes the reporting display to not show calculation data when reporting on the 30 second price update display.
This will remain this way, till I or someone else implments the promise pattern on the calculate function of the strategies.

All the strategies are configured to use a 1000 period look-back when calculating (they will start calculating as soon as the minimum period is reached regardless of not having 1000 periods), this may effect memory limited devices like a Pi and may need to be adjusted when using. This is only a possibility and would need to be tested.    I do not have a Pi to test on,  but I could see this as being a concern.

JohnnyVicious:
Ah nice, I see you added the current close for the calculation 👍

DeviaVir (approved):
:+1: great work, let's see it in the wild

PR #1624, opened 2018-06-18
Debug lib

defkev:
Move debugging to its own library so it can be (re)used in every module instead of just engine.js

Furthermore this will expose the current debugging state (in debug.on) globally, use it instead of declaring so.debug locally in each module from process.argv

To use it first import `lib/debug.js` and then call `debug.msg()`
The current debugging state is available in `debug.on` which is either **true** or **false** depending on `--debug`, using the `L` key will the bot is running or by calling `debug.flip()`

Example:
```js
const debug = require('lib/debug') // change path depending on where you are

module.exports = function foo() {
  debug.msg('bar')

  if (debug.on) {
    debug.flip() // will disable debugging
  }
}
```

### DIFF
--- a/boot.js
+++ b/boot.js
@@ -11,6 +11,8 @@ module.exports = function (cb) {
   var config = {}
   var overrides = {}
 
+  module.exports.debug = args.debug
+
   // 1. load conf overrides file if present
   if(!_.isUndefined(args.conf)){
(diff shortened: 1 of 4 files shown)


DeviaVir (approved):
:+1:

PR #1626, opened 2018-06-18
CEXIO: Use debug lib

defkev:
Might as well use it :smirk:


station384:
EDIT:   @defkev  I sent a pull request to your branch with the fixes so it will pass the .eslint tests

There are some eslint errors 
Part of your code is using windows style CRLF instead of unix style LF as defined in .eslint (I would point out the individual lines but it seems to be almost all)

I only bring this up because when I tried to merge in your changes into one of my branches I got a git failure, along with tab spacing, and undefined variables.  (all of which does not effect running of the code so this feels kinda "nit-picking")

defkev:
Guilty as charged :grin: 

Dunno why Travis passed on exchange.js though it is still set to Windows/CRLF instead of Unix/LF on my box.

Meanwhile #1624 failed initially till i converted it in 460cd933a99df7c9538f4f7a3994d16a6be24b8a 

Odd :thinking:

DeviaVir (approved):
:+1: sweet

PR #1631, opened 2018-06-21
Add minimum trades and fitnessCalcType options

station384:
--minTrades will try and achieve the minimum x amount of trades before using normal fitness calculations
--fitnessCalcType allows for different fitness priorities.  i.e. profit does not care how many wins or losses occur.  it only looks at how much of a return is occurring.   wl does not care about profit, only more wins than losses, profitwl tries to get the best profit, while taking into consideration wins and losses, but rates profit higher.  classic uses the legacy calculation and is the default.

Addition modifications
Add PROPERTY_RANDOM_CHANCE to phenotype process to allow for injection of entirely new species. this helps prevent stagnation during a long run.

At some point floatScanWindow was removed from backtester.  Commented out code in darwin.js that applies to this change

### DIFF
--- a/commands/sim.js
+++ b/commands/sim.js
@@ -158,6 +158,10 @@ module.exports = function (program, conf) {
           output_lines.push('error rate: ' + (sells ? n(losses).divide(sells).format('0.00%') : '0.00%').yellow)
         }
         options_output.simresults.start_capital = s.start_capital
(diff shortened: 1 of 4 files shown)


DeviaVir (approved):
:+1: thanks for keeping this in sync

PR #1656, opened 2018-07-09
Improvement/typescript refactor

notVitaliy:
See https://github.com/DeviaVir/zenbot/issues/1652


GarethWright:
Bloody good work! This looks leaps and bounds better

DeviaVir:
Thinking on creating a zenbot.ts and renaming this repo to zenbot.js and seeing which one gets more traction and further development to nominate that one as the default for `/zenbot`. Thoughts? My idea behind it is that I would like to prevent this project from dying because folks do not want to work with typescript.

.gitignore
```diff
@@ -39,3 +39,4 @@ gen.*.bat
 gen.*.sh
 .env
 scripts/auto_backtester/backtesting_*.csv
+.vscode
```
DeviaVir, line 42: I don't mind, but you could also add this to your global gitignore locally.

GarethWright: Splitting them out kinda makes sense but why not a separate branch? Easier to cherry pick changes applicable to both?

GarethWright:
@DeviaVir Splitting them out kinda makes sense but why not a separate branch? Easier to cherry pick changes applicable to both?

notVitaliy:
Already discussed with @DeviaVir on discord, but just wanted to put this down here as well.

I'm thinking a new branch would be the best way to go. A new repo will, for one, be at a disadvantage since all the stars and followers will be reset; it will create the need to maintain feature parity between the two projects which will inevitably succumb to piles of technical debt.

In case there's a huge backlash against typescript there's always the option of transpiling the source to plain JS. All we'd have to do is run `tsc` from the command line and commit the generated files. From there, we could work on the plain JS files. As long as we don't compile down to an ES5 target, the transpiled code won't look like a nightmare.

Ultimately, I want this project to be badass. There's huge potential here. I don't have a dogmatic preference on which language is used, I just think TS would be a huge benefit here and it'll become clear once I'm done with this refactoring.

p.s. I've started work on state management last night and re-architecting the data schema. This change is going to break most backward compatibility (unless we create a migration script) but it introduces an immutable and observable state tree, which is nice especially in comparison to passing around the `s` object.

vertic4l:
@notVitaliy Nice work, i'm working with TypeScript as well. Is the TS Version up2date? Can we progress on this ?

dawez:
@notVitaliy I also think that the move to TS will bring some new life in the project. As per today TS is the standard in the JS world.

Do you have any update on this? I notice as per today this branch is 688 commits behind unstable.
Probably most of them are just exchange updates.

notVitaliy:
@dawez I actually turned this PR into a forked repo and then detached and completely re-wrote the whole bot. It's [Magic8Bot](https://github.com/magic8bot/magic8bot).

I was pretty active in developing it for a few months but life got in the way.

vertic4l:
@notVitaliy Sick! I will have a look at it and hopefully contribute if possible. Let's keep in contact!

Issue #1647, opened 2018-07-04
Binance API cannot be used to trade. Error: binance {"code":-1021,"msg":"Timestamp for this request is outside of the recvWindow."


coolcode:
### System information
- **Have I written custom code (as opposed to using zenbot vanilla)**: No.
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Mac OS El Capitan
- **Zenbot version** (commit ref, or version): Commits on Jul 1, 2018
- **Zenbot branch**: unstable
- **NodeJS version**: v9.11.1
- **Python version (when using a python script)**: Not using python
- **Exact command to reproduce (include everything)**: No
- **Did I make any changes to conf-sample.js?**: 
c.selector = 'binance.ETH-USDT' 
c.strategy = 'trend_ema'

### Describe the problem
Binance API cannot be used to trade.

### Source code / Error logs
Binance API is down! unable to call getBalance, retrying in 20s
[ { currency: 'USDT', asset: 'ETH' } ]
An error occurred { Error: binance {"code":-1021,"msg":"Timestamp for this request is outside of the recvWindow."}
    at binance.handleErrors (/Users/brucelee/projects/eth/zenbot/node_modules/ccxt/js/binance.js:863:31)
    at response.text.then (/Users/brucelee/projects/eth/zenbot/node_modules/ccxt/js/base/Exchange.js:513:18)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:182:7) constructor: [Function: InvalidNonce] }

mrseeker:
Is your systemdate correct? Do you have an NTP client running?

coolcode:
It's correct. NTP is default running in MacOS, which is using 'time.apple.com'.

jraviotta:
I have the same issue on both Windows and Mac with zenbot on Docker.

```
Binance API is down! unable to call getBalance, retrying in 20s
[ { currency: 'USDT', asset: 'EOS' } ]
An error occurred { Error: binance {"code":-1021,"msg":"Timestamp for this request was 1000ms ahead of the server's time."}
    at binance.handleErrors (/app/node_modules/ccxt/js/binance.js:863:31)
    at response.text.then (/app/node_modules/ccxt/js/base/Exchange.js:513:18)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7) constructor: [Function: InvalidNonce] }
```
[...]

PR #1659, opened 2018-07-12
Time-sync function to update container time

jraviotta:
Time in Docker containers on Mac drifts away from system time. This adds a target to `Makefile` which forces the container time to update.
#1647

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       - ./:/app/
       - /app/node_modules/
       - /app/dist/
-      # Avoid syncing database to our server
+     # Avoid syncing database to our server
       - /app/database/
     restart: always
     tty: true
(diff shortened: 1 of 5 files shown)


docker-compose.yml
```diff
@@ -8,21 +8,22 @@ services:
...
     depends_on:
       - mongodb
     environment:
       - MONGODB_PORT_27017_TCP_ADDR=mongodb
-
   mongodb:
     image: mongo:latest
     volumes:
-      - ./database:/data/db
+      - data-volume:/data/db
```
DeviaVir: I've had this PR'ed a few times, but I'm not sure if I like it. The data is somewhere hidden in docker caches vs just in the same directory.

jraviotta:
I think I removed the questionable lines and updated the PR. It should now include only the change to the Makefile.

DeviaVir (approved):

PR #1660, opened 2018-07-12
Cross-platform Makefile

jraviotta:
This adds variables and logic to `Makefile` for cross-platform use.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,53 @@
+# Check if this is Windows
+ifneq (,$(findstring WINDOWS,$(PATH)))
+WINDOWS := True
+endif
+
+# Set shell to cmd on windows
+ifdef WINDOWS
+SHELL := C:/Windows/System32/cmd.exe
+endif
+
+# Don't use sudo on windows
+SUDO := "sudo"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -9,68 +59,69 @@ list:
...
-	sudo docker-compose rm --force server
-	sudo docker-compose rm --force mongodb
-	sudo docker-compose rm --force adminmongo
-	sudo docker-compose build --no-cache
-	sudo docker-compose up -d --force-recreate
+	$(SUDO) docker-compose stop
+	$(SUDO) docker-compose pull
+	$(SUDO) docker-compose rm --force server
+	$(SUDO) docker-compose rm --force mongodb
+	-$(SUDO) docker-compose rm --force adminmongo
```
kenorb, line 112: `-`?

jraviotta: Are you asking about line 112?
That makes the command optional and prevents the target from stopping if it fails. If the user is using adminmongo, the command executes, otherwise the container doesn't exist so the target will report that the command failed and move on.

DeviaVir (approved):
:+1: great

PR #1686, opened 2018-08-23
 Fixed max_size for ADABTC pair

cvinacua:
On pair ADABTC max_size was too low, so selling operation was failing.


DeviaVir:
You'll need to implement a resolution for that in the `update-script.sh`: https://github.com/DeviaVir/zenbot/blob/unstable/extensions/exchanges/binance/update-products.sh

Otherwise these changes will just be blown away the next time our automated updater runs.

cvinacua:
Thanks, now I understand that the products file is generated automatically. My apologies for bothering you with this silly thing

PR #1690, opened 2018-08-24
 Fixed max_size. The script was taking maxPrice instead of maxQty.

cvinacua:
I have fixed the update-products.sh which was using the wrong parameter to set the max_size


DeviaVir:
Conflicted, sorry. You may want to leave out the changes in products.json in your commits.

PR #1738, opened 2018-10-08
Set default value for min_periods in trade command

BAKFR:
Using paper mode, the `min_periods` option was used without being defined. It would lead to date calculation returning `NaN`, and thus provoking errors (at least on Binance exchange).

This commit fixes this.

### DIFF
--- a/commands/trade.js
+++ b/commands/trade.js
@@ -87,6 +87,7 @@ module.exports = function (program, conf) {
         console.log(('--buy_max_amt is deprecated, use --deposit instead!\n').red)
         so.deposit = so.buy_max_amt
       }
+      if (!so.min_periods) so.min_periods = 1
 
       so.selector = objectifySelector(selector || conf.selector)      
       var engine = engineFactory(s, conf)


DeviaVir (approved):
:+1:

olexiyb:
This change actually impacting for example sar strategy and when we run paper trading it's failing with 

C:\Users\olexi\WebstormProjects\zenbot-web\packages\zenbot-api\zenbot\extensions\strategies\sar\strategy.js:32
          s.sar = Math.max(s.lookback[1].high, s.lookback[0].high)
                                         ^

TypeError: Cannot read property 'high' of undefined
    at Object.calculate (C:\Users\olexi\WebstormProjects\zenbot-web\packages\zenbot-api\zenbot\extensions\strategies\sar\strategy.js:32:42)

Which strategy actually was failing?

BAKFR:
@olexiyb I don't remember, but I've tested several strategies that was failing.

Basically, before this commit, `min_periods` was used several times (in `trades.js`) without it being initialized. This was causing an error to all strategies not requiring `min_periods` has a strategy option.

Can you try to call the sar strategy with the `min_periods` defined manually ? It's just a guess, but I thing it could be explained by the default value not being set at 52 (as requested in the sar strategy), thus creating only one period in `s.lookback`.

**Edit:** Feel free to mentions me in another issue instead of commenting in this one.

PR #1742, opened 2018-10-16
Fix simulation fees

BAKFR:
The commit 1dd7955 is a simplification of the code of `processTrade()`. It bring no change besides having few lines.

The commit 2f0be4f rewrites most of `processBuy()` and `processSell()`. The fee calculation were unnecessarily complicated.
Beside, `processSell()` was computing fees as a percent of *asset*, but was removing them from balance as *currency*, leading to incorrect values.

I'm not sure it's the only bug in simulation mode, but zenbot not longer reports me a balance of -120BTC, so that's a good sign.


DeviaVir (approved):
:+1: neat

PR #1754, opened 2018-10-30
notifier filter for completed trades

MCYBA:
when ı use notifer for seing my trades data, canceled orders become so big waste, then ı create a filter to zenbot for sending notifier only completed messages.

### DIFF
--- a/commands/trade.js
+++ b/commands/trade.js
@@ -57,6 +57,7 @@ module.exports = function (program, conf) {
     .option('--use_fee_asset', 'Using separated asset to pay for fees. Such as binance\'s BNB or Huobi\'s HT', Boolean, false)
     .option('--run_for <minutes>', 'Execute for a period of minutes then exit with status 0', String, null)
     .option('--debug', 'output detailed debug info')
+    .option('--only_completed_trades', 'Message filter for only completed trades to notifier.', Boolean, conf.notifiers.only_completed_trades)
     .action(function (selector, cmd) {
       var raw_opts = minimist(process.argv)
       var s = {options: JSON.parse(JSON.stringify(raw_opts))}
@@ -76,6 +77,7 @@ module.exports = function (program, conf) {
           so[k] = cmd[k]
         }
       })
+      so.only_completed_trades = (cmd.only_completed_trades || conf.only_completed_trades)
       so.currency_increment = cmd.currency_increment
       so.keep_lookback_periods = cmd.keep_lookback_periods
       so.use_prev_trades = (cmd.use_prev_trades||conf.use_prev_trades)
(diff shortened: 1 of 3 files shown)


commands/trade.js
```diff
@@ -57,6 +57,7 @@ module.exports = function (program, conf) {
     .option('--use_fee_asset', 'Using separated asset to pay for fees. Such as binance\'s BNB or Huobi\'s HT', Boolean, false)
     .option('--run_for <minutes>', 'Execute for a period of minutes then exit with status 0', String, null)
     .option('--debug', 'output detailed debug info')
+    .option('--only_completed_trades', 'Message filter for only completed trades to notifier.', Boolean, conf.notifiers.only_completed_trades)
```
DeviaVir: I don't think we should add this, but just keep it as a `conf` option instead.

lib/engine.js
```diff
@@ -425,8 +425,12 @@ module.exports = function (s, conf) {
               }, conf.wait_for_settlement)
             }
             else {
-              pushMessage('Buying ' + formatAsset(size, s.asset) + ' on ' + s.exchange.name.toUpperCase(), 'placing buy order at ' + formatCurrency(price, s.currency) + ', ' + formatCurrency(quote.bid - Number(price), s.currency) + ' under best bid\n')
-              doOrder()
+              if(!so.only_completed_trades){
```
DeviaVir: this should probably be 
```suggestion
              if(!conf.notifiers.only_completed_trades){
```

lib/engine.js
```diff
@@ -473,8 +477,12 @@ module.exports = function (s, conf) {
               }, conf.wait_for_settlement)
             }
             else {
-              pushMessage('Selling ' + formatAsset(size, s.asset) + ' on ' + s.exchange.name.toUpperCase(), 'placing sell order at ' + formatCurrency(price, s.currency) + ', ' + formatCurrency(Number(price) - quote.bid, s.currency) + ' over best ask\n')
-              doOrder()
+              if(!so.only_completed_trades){
```
DeviaVir:
```suggestion
              if(!conf.notifiers.only_completed_trades){
```

commands/trade.js
```diff
@@ -76,6 +76,7 @@ module.exports = function (program, conf) {
           so[k] = cmd[k]
         }
       })
+      so.only_completed_trades = conf.only_completed_trades
```
DeviaVir: I think you can leave this out entirely.
```suggestion
      
```

DeviaVir (changes requested):
some more things

MCYBA:
Mister, sorry about i taken your time. i don't understand, there is no problem on my repository. Could you explain me about this case?

DeviaVir (approved):

PR #1757, opened 2018-10-31
* removing binance VEN as it was renamed to VET and no VEN anymore on binance

olexiyb:


DeviaVir:
This file is auto-generated on a daily basis, you may want to look into why this product is being returned from https://github.com/DeviaVir/zenbot/blob/1b3a3e8a17b709c847125c0c42f5147eb5cbbb01/extensions/exchanges/binance/update-products.sh

PR #1763, opened 2018-11-11
small logic cleanup

olexiyb:
* added else to avoid confusion with previous if statement as so.order_type can not change in previous if
* removed redundant duplicate check of s.last_day

### DIFF
--- a/lib/engine.js
+++ b/lib/engine.js
@@ -542,8 +542,7 @@ module.exports = function (s, conf) {
         if (s.exchange.makerFee) {
           fee = n(s.buy_order.size).multiply(s.exchange.makerFee / 100).value()
         }
-      }
-      if (so.order_type === 'taker') {
+      } else if (so.order_type === 'taker') {
         if (s.exchange.takerFee) {
           fee = n(s.buy_order.size).multiply(s.exchange.takerFee / 100).value()
         }
@@ -908,7 +907,7 @@ module.exports = function (s, conf) {
       return
     }
     var day = (new Date(trade.time)).getDate()
-    if (s.last_day && s.last_day && day !== s.last_day) {
+    if (s.last_day && day !== s.last_day) {
       s.day_count++
     }
     s.last_day = day


lib/engine.js
```diff
@@ -542,8 +542,7 @@ module.exports = function (s, conf) {
         if (s.exchange.makerFee) {
           fee = n(s.buy_order.size).multiply(s.exchange.makerFee / 100).value()
         }
-      }
-      if (so.order_type === 'taker') {
+      } else if (so.order_type === 'taker') {
```
DeviaVir, line 545: I don't use `else if`, but maybe my knowledge is outdated?
https://dev.to/genta/theres-no-else-if-in-js--24f9

olexiyb: https://www.w3schools.com/js/js_if_else.asp

The else if Statement
Use the else if statement to specify a new condition if the first condition is false

trivalik: The "else if" construct makes, in case of no optimization, that the condition in "else if" is not evaluated if the "if" statement was above already true.

olexiyb: I would not rely on optimization, I just know that with else we will skip condition check (so.order_type === 'taker') for case when (so.order_type === 'maker'), without else this will always check (so.order_type === 'taker').
In general this will not drastically improve performance, but usually just good practice to use else if

olexiyb: BTW, the author of article https://dev.to/genta/theres-no-else-if-in-js--24f9
discussing problem that does not make any sense to me.
With braces or without, it does not matter, code will work as expected and result will be the same.
For example I can start to tell that 
(a - b + c) 
is wrong we have to use
(a + c - b) 
This is the same!

trivalik: I guess he wants to say that
`
else if (xxx) {
}
`
and
`
else { if(xxx){
}
}
`

do the same, but when you forget a { or } you fucked up. Anyway it's just to know that "else if" use the else without brackets, which is for some code guides a code smell.

--> back to my answer, the different to a missing else is that every if has to be checked, in case the optimizer does not detect that, because if first if is true, the else will never be looked at

PR #1764, opened 2018-11-11
improve bollinger strategies

olexiyb:
* tiny performance improvement to skip calculation of bollinger bands for each trade.
see lib/bollinger.js
line with
if (key in s.period) return 
check

* keep only bands itself to decrease size of data on 2 weeks sample file size 
201 Kb vs 114 Kb

var data = [{"period_id":"h428313","size":"1h","time":1541926800000,"open":213.53,"high":213.58,"low":213.49,"close":213.49,"volume":466.78638,"close_time":1541930399999,"latest_trade_time":1541927220000,"bollinger":{"upper":[null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,215.41419216485025],"mid":[null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,213.73649999999998]

after the change:

var data = [{"period_id":"h428313","size":"1h","time":1541926800000,"open":213.53,"high":213.58,"low":213.49,"close":213.49,"volume":466.78638,"close_time":1541930399999,"latest_trade_time":1541927220000,"bollinger":{"upperBound":215.41419216485025,"midBound":213.73649999999998,"lowerBound":212.0588078351497}}

### DIFF
--- a/lib/bollinger.js
+++ b/lib/bollinger.js
@@ -1,14 +1,26 @@
 // Bollinger Bands
 var bollingerbands = require('bollinger-bands')
+
 module.exports = function bollinger (s, key, length, source_key) {
   if (!source_key) source_key = 'close'
   if (s.lookback.length > length) {
+    // skip calculation if result already presented as we use historical data only,
+    // no need to recalculate for each individual trade
+    if (key in s.period) return
     let data = []
     for (var i=length-1; i>=0; i--) {
       data.push(s.lookback[i][source_key])
     }
-    let result = bollingerbands(data, length, s.options.bollinger_time)
-    s.period[key] = result
+    const result = bollingerbands(data, length, s.options.bollinger_time)
+    const upperBound = result.upper[result.upper.length-1]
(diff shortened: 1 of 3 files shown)


lib/bollinger.js
```diff
@@ -1,14 +1,26 @@
 // Bollinger Bands
 var bollingerbands = require('bollinger-bands')
+
 module.exports = function bollinger (s, key, length, source_key) {
   if (!source_key) source_key = 'close'
   if (s.lookback.length > length) {
+    // skip calculation if result already presented as we use historical data only,
+    // no need to recalculate for each individual trade
+    if (key in s.period) return
```
olexiyb, line 9: This is performance improvement. As I found we call this for each individual trade, but if bands calculated we do not need to recalculate them

olexiyb:
performance_diff

DeviaVir (approved):
:+1: thanks for contributing

PR #1768, opened 2018-11-17
fix mongo connect warning

olexiyb:
(node:121) DeprecationWarning: current URL string parser is deprecated, and will be removed in a future version. To use the new parser, pass option { useNewUrlParser: true } to MongoClient.connect


DeviaVir (approved):
:+1:

PR #1769, opened 2018-11-17
fix collection.ensureIndex is deprecated. Use createIndexes instead.

olexiyb:
(node:6880) DeprecationWarning: collection.ensureIndex is deprecated. Use createIndexes instead.

see https://docs.mongodb.com/manual/reference/method/db.collection.ensureIndex/


DeviaVir (approved):
:+1:

PR #1773, opened 2018-11-19
avg_slippage_pct should be used for taker orders

olexiyb:
It looks like original code had
https://github.com/DeviaVir/zenbot/issues/1109#issuecomment-357411767

but later this changed

https://github.com/DeviaVir/zenbot/pull/1378

@DeviaVir , @glennfu , @krystophv  can you guys review my fix. I feel this can be important.

### DIFF
--- a/extensions/exchanges/sim/exchange.js
+++ b/extensions/exchanges/sim/exchange.js
@@ -202,7 +202,7 @@ module.exports = function sim (conf, s) {
     let price = buy_order.price
 
     // Add estimated slippage to price
-    if (so.order_type === 'maker') {
+    if (so.order_type === 'taker') {
       price = n(price).add(n(price).multiply(so.avg_slippage_pct / 100)).format('0.00000000')
     }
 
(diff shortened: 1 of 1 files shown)


DeviaVir:
I'm wondering if both taker and maker don't have to deal with slippage to some extent? I'd guess a taker more than a maker so your PR makes sense.

mrseeker:
Taker orders are "market" orders. They should not have slippage at all, unless you have a very large bulky order that causes the price to drop. Makers however are more affected by slippage, since they start at a certain point and with high volatility can get a lot of slippage.

https://www.investopedia.com/terms/s/slippage.asp

kztd:
slipppage is the idea that an order may get filled at a price worse than expected. This should never happen on a limit order and normally that is a 'maker' order.  Taker orders are generally market orders and the expected price is the inside market at the time the order is placed. There are several reasons a market order may be executed a a price worse than expected, e.g. if the if the inside market size is less than the order size, or if the market moves away between the time the order is sent and when it is executed.

Issue #1777, opened 2018-11-22
Bitfinex issue


leozzo:
### System information
- **Have I written custom code (as opposed to using zenbot vanilla)**: NO
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu 16 and 18
- **Zenbot version** (commit ref, or version): last at 22/nov/2018
- **Zenbot branch**: any
- **NodeJS version**: 8 and 9
- **Python version (when using a python script)**:
- **Exact command to reproduce (include everything)**:  any command, also the same in your example
- **Did I make any changes to conf-sample.js?**: NO

### Describe the problem
The zenbot doesn't work anymore with Bitfinex since October 2018+ version.
Since API 2.0 was implemented (in the zenbot) it doesnt work anymore, give tons of errors.
Please can you fix it? Thank you

felpasl:
same here, for last version, --paper works, but not able to get balance to start real trade

Luehrsen:
I have spent some time debugging this. This is what I've found so far: 

---

https://github.com/DeviaVir/zenbot/blob/6e855d06099962f90fabfbf1193385e43de307f9/extensions/exchanges/bitfinex/exchange.js#L165

The if-statement above is never false, therefore the `ws.open()` is never called. `ws_connecting` is set to true here, but the whole variable does not make much sense to me. 

https://github.com/DeviaVir/zenbot/blob/6e855d06099962f90fabfbf1193385e43de307f9/extensions/exchanges/bitfinex/exchange.js#L222

---

https://github.com/DeviaVir/zenbot/blob/6e855d06099962f90fabfbf1193385e43de307f9/extensions/exchanges/bitfinex/exchange.js#L235

The `ws` and `wu` events are never fired. (But they do arrive on the `message` event.)

--- 

So I believe that v2.0 is less compatible with v1.x than the Bitfinex repo states. So we either have to rewrite the `exchange.js` or downgrade to v1.2.1.

anon1335:
Hi, thanks for exploring this @Luehrsen , is there any simple way to downgrade to v1.2.1?

Issue #1734, opened 2018-10-07
Update done automatically by greenkeeper broke Bitfinex support.


redknightlois:
### System information
- Yes, custom strategies
- Windows and Linux
- latest
- Unstable

### Describe the problem
Bitfinex update done automatically by greenkeeper broke the exchange support. 
https://github.com/DeviaVir/zenbot/commit/348650579b5749cb6ef13f9de4a7f89385aeb211
The new library (v2.0) is subtly different. 

### Source code / Error logs
```
server_1        | TypeError: Cannot read property 'constructor' of null
server_1        |     at new BFX (/app/node_modules/bitfinex-api-node/index.js:31:14)
server_1        |     at publicClient (/app/extensions/exchanges/bitfinex/exchange.js:26:41)
server_1        |     at Object.getTrades (/app/extensions/exchanges/bitfinex/exchange.js:441:22)
server_1        |     at getNext (/app/commands/backfill.js:82:18)
server_1        |     at /app/commands/backfill.js:66:9
server_1        |     at result (/app/node_modules/mongodb/lib/utils.js:414:17)
server_1        |     at executeCallback (/app/node_modules/mongodb/lib/utils.js:406:9)
server_1        |     at handleCallback (/app/node_modules/mongodb/lib/utils.js:128:55)
server_1        |     at cursor.close (/app/node_modules/mongodb/lib/operations/cursor_ops.js:211:62)
server_1        |     at handleCallback (/app/node_modules/mongodb/lib/utils.js:128:55)
server_1        |     at completeClose (/app/node_modules/mongodb/lib/cursor.js:893:14)
server_1        |     at _endSession (/app/node_modules/mongodb/lib/cursor.js:904:37)
server_1        |     at Cursor._endSession (/app/node_modules/mongodb-core/lib/cursor.js:194:5)
server_1        |     at Cursor._endSession (/app/node_modules/mongodb/lib/cursor.js:226:59)
server_1        |     at Cursor.close (/app/node_modules/mongodb/lib/cursor.js:904:19)
server_1        |     at cursor._next (/app/node_modules/mongodb/lib/operations/cursor_ops.js:211:23)
server_1        |     at handleCallback (/app/node_modules/mongodb-core/lib/cursor.js:203:5)
server_1        |     at _setCursorNotifiedImpl (/app/node_modules/mongodb-core/lib/cursor.js:561:38)
[...]

PR #1778, opened 2018-11-23
bitfinex backfill is broken

olexiyb:
This is first part to fix backfill functionality only

I have used rate limit recommendation

https://bitcoin.stackexchange.com/questions/36952/bitfinex-api-limit

with some modifications as I found that 1 request per second is too often for bitfinex so I use 2000 ms timeout to retrieve bitfinex trades.


olexiyb:
this should partially help with #1777 and #1734 issues

DeviaVir (approved):
:+1:

Issue #1775, opened 2018-11-21
Multiple Bots - same HTML file


mroxso:
### System information
- **Have I written custom code (as opposed to using zenbot vanilla)**: no
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Linux/Debian 9
- **Zenbot version** (commit ref, or version): latest
- **Zenbot branch**: unstable
- **NodeJS version**: v10.13.0
- **Python version (when using a python script)**: n/a
- **Exact command to reproduce (include everything)**: zenbot trade --paper --conf my_conf.js
- **Did I make any changes to conf-sample.js?**: I copied the conf-sample.js twice, for each bot another conf file. But I did not change the original.

I ran into a problem while running multiple zenbots on one system.
The problem is, that 2 bots are using the same name for their HTML statistics dump file.

Is there already a feature to manually set the name of the HTML file?
How is the filename generated?

DJCrashdummy:
well for `zenbot sim ...` there is a `--filename`-option you may also try with `zenbot trade ...`, but probably it won't work.

it seems that the names are generated from the timestamps of the last tick or something like that... as a workaround you can try to not stop both instances at the same time.

all in all you are right, if it doesn't work already, a `--filename`-option also for the `zenbot trade ...`-command would definitively make sense!

mroxso:
> all in all you are right, if it doesn't work already, a `--filename`-option also for the `zenbot trade ...`-command would definitively make sense!

Thank you! I tried it and it should also work with trade command!

PR #1779, opened 2018-11-23
added --filename to trade --help

DJCrashdummy:
according to #1775 the possibility to define a filename is not listed in `trade --help`, but works... this probably avoids future questions.

---

hopefully this commit is correct, because i sadly have hardly any idea of js!?


mroxso:
But is ""none" to disable" correct? Because when you not specify this command, it will generate a name for it, doesn't it?

DJCrashdummy:
i've tested it (for `zenbot sim ...` and `zenbot trade ...`) and it is correct:

- if you don't specify any name, the default name is taken
- if you enter "`none`" (therefore the quotes) as name, no file is created

mroxso:
Oh okay alright :)

DeviaVir (approved):

PR #1790, opened 2018-12-04
Binance: use real trades instead of OHLCV when asking for trades

BAKFR:
In the `getTrades()` method, Binance was previously fetching OHLCV data, in order to generate virtual trades.
The returned result was composed of one (fake) trade per minute, with the direction and volume of the OHLCV data.
Beside, a large part of code was unused and/or contained obvious errors (use of undefined variables, ...)

This commit rewrites the `getTrades()` method to use the real trades returned by Binance.

----
Also, when I get no trade and the `startTime` is older than a hour, I use `fetchOHLCV()` for confirmation. Without that, the backfill would stop when there are some holes in the trades (by example when Binance is in maintenance).

### DIFF
--- a/extensions/exchanges/binance/exchange.js
+++ b/extensions/exchanges/binance/exchange.js
@@ -1,16 +1,14 @@
 const ccxt = require('ccxt')
-, path = require('path')
-// eslint-disable-next-line no-unused-vars
-, colors = require('colors')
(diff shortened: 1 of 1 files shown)


DeviaVir (approved):
Looks good to me, thanks for contributing.

PR #1795, opened 2018-12-10
Fix binance product update

BAKFR:
It was far easier than I though: Poloniex already had the *notional* thing of Binance, but it was called min_total.

I've not tested to execute an order, but I've read it twice. It should not have any problem.
There are more details in the commit messages.

### DIFF
--- a/lib/engine.js
+++ b/lib/engine.js
@@ -305,6 +305,14 @@ module.exports = function (s, conf) {
     })
   }
 
+  function isOrderTooSmall(product, quantity, price) {
+    if (product.min_size && Number(quantity) < Number(product.min_size))
+      return true
+    if (product.min_total && n(quantity).multiply(price).value() < Number(product.min_total))
+      return true
+    return false
+  }
+
   // if s.signal
   // 1. sync balance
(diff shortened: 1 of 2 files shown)


lib/engine.js
```diff
@@ -305,6 +305,14 @@ module.exports = function (s, conf) {
     })
   }
 
+  function isOrderTooSmall(product, quantity, price) {
+    if (product.min_size && Number(quantity) < Number(product.min_size))
```
DeviaVir, line 309: you may want to add a check if `product` has `min_size` and `min_total`, that could prevent a crash here.

DeviaVir (approved):

lib/engine.js
BAKFR: There is no need, since `product.min_size` will resolve to `undefined` if it does not exists (and thus the condition will fail).
The only possible crash would be if `product` was `null` or `undefined`

DeviaVir: ugh of course, you'd only check if you're looking inside the object, you're totally right.

PR #1800, opened 2018-12-21
Add host listening all host in container

julesGoullee:
When you update adminmongo now it's listening only on localhost by default when you launch in a container you cannot connect from outside, I add all interface listening in the compose file


DeviaVir:
I think exposing to public by default is a security leak, we should set it to `127.0.0.1` and document how to expose it to world by replacing it to `0.0.0.0`. Thoughts?

mroxso:
If so, i would prefer DeviaVir's method.

julesGoullee:
The point is not exposed to the world only expose to outside the container, (eg local host of the container)

DeviaVir:
If you'd add:
```
ports:
  - 127.0.0.1:1234:1234
``` 
I would agree

julesGoullee:
Yep better

PR #1805, opened 2018-12-28
Changes in Ichimoku Score

mrseeker:
Two fixes: One is removing a duplicate of the ichimoku score strategy.
Second one is adding some changes to make the strategy perform better. There were some issues with a bot getting stuck when the score went to exactly 50, and there were some issues when the score went between the buy and sell level (neutral zone) and back up...


DeviaVir (approved):
:+1:

Issue #1801, opened 2018-12-27
Partial order can cause volume to go NaN


mrseeker:
### System information
- **Have I written custom code (as opposed to using zenbot vanilla)**: Yes
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Linux Debian 9
- **Zenbot version** (commit ref, or version):  f23ae68
- **Zenbot branch**: master
- **NodeJS version**: v10.6.0
- **Python version (when using a python script)**: n/a
- **Exact command to reproduce (include everything)**: ./zenbot.sh trade
- **Did I make any changes to conf-sample.js?**: Yes

### Describe the problem
I woke up to see the bot do a "NaN" order, this is the telegram version:
"Selling NaN BTC on GDAX: placing sell order at 3320.01 EUR, 0.0100000 EUR over best ask"

Apparently it did a partial order that got filled at the exact moment that the bot tried to cancel the last order. The order did fill, but a race condition occurred or the exchange was down, causing the bot to fill in an order of "NaN" BTC. This in return caused the bot to get into a loop of death since the order could never get filled...

### Source code / Error logs
Most likely source is lib/engine.js, placeOrder getting a NaN order in. It should not execute the order if the size is not higher than 0 or invalid...

Fix is really simple: Check for NaN, if true then drop order as "invalid".

PR #1806, opened 2018-12-28
Fix for NaN issue #1801

mrseeker:
Fix for NaN issue #1801


DeviaVir (approved):
:+1: thanks for the contribution!

Issue #1809, opened 2019-01-02
Renko strategy error


hyuce:
### System information
- **Have I written custom code (as opposed to using zenbot vanilla)**: zenbot unstable (Dec 31, 2018)
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Linux Mint 19.1 Tessa
- **Zenbot version** (commit ref, or version): 4.1.0 (unstable)
- **Zenbot branch**: unstable
- **NodeJS version**: 8.15.0
- **Python version (when using a python script)**: 
- **Exact command to reproduce (include everything)**: `./zenbot.sh sim binance.ETH-PAX --strategy=renko --min_periods=200`
- **Did I make any changes to conf-sample.js?**: no

### Describe the problem
When I use renko strategy with sim, I get error in following lines.
### Source code / Error logs
```console
./zenbot.sh sim binance.ETH-PAX --strategy=renko --min_periods=200 
/home/username/zenbot/extensions/strategies/renko/strategy.js:16
    prclose = Math.round(s.lookback[s.options.min_periods].close)
                                                           ^

TypeError: Cannot read property 'close' of undefined
    at Object.calculate (/home/username/zenbot/extensions/strategies/renko/strategy.js:16:60)
    at updatePeriod (/home/username/zenbot/lib/engine.js:162:16)
    at withOnPeriod (/home/username/zenbot/lib/engine.js:772:5)
    at onTrade (/home/username/zenbot/lib/engine.js:952:7)
    at /home/username/zenbot/lib/engine.js:913:5
    at /home/username/zenbot/node_modules/async/dist/async.js:4096:9
    at Object.process (/home/username/zenbot/node_modules/async/dist/async.js:2344:17)
    at /home/username/zenbot/node_modules/async/dist/async.js:2252:19
    at Immediate._onImmediate (/home/username/zenbot/node_modules/async/dist/async.js:119:16)
    at runCallback (timers.js:810:20)

DeviaVir:
fixed by https://github.com/DeviaVir/zenbot/pull/1810

PR #1810, opened 2019-01-02
Fix for Renko strategy error #1809

qubicle:
Execute calculation only if lookback data exists


DeviaVir (approved):

sweepglass:
cols.push(z(8, n(rclose), ' '))
                       ^
ReferenceError: rclose is not defined

PR #1816, opened 2019-01-16
WIP: Rework docs

DJCrashdummy:
1. Hopefully I assumed right and you don't use the wiki (although it is active), but a docs-folder with md-files?

1. This PR is **by far NOT done** yet! - I just wanted to get in touch (and eventually feedback), to see if my ideas/changes are welcome, before I put too much time and effort in it.

1. I want to rework the docs, which means...
    - In first place shorten the main readme, because it almost overwhelms the reader at the first impression (and also as user of zenbot it isn't really a pleasure to handle it).
    - Create single and shorter files grouped by topic in the docs-folder (as it is already done for the `exchanges`).
    - Create a kind of an index-readme at the docs-folder.
    - And of course update some outdated parts.

PR #1820, opened 2019-01-23
[WIP] Upgrade 

BAKFR:
**Do not merge: it's a work in progress**

Following the conversation in #1816 

According to https://github.com/nodejs/Release#release-schedule
The v8 branch will be supported until December 2019. It's the default version available on Ubuntu (without ppa).
The v9 (used by Travis) isn't supported anymore.
The v10 is the last LTS, and the v12 is the current stable version.

I believe we should set the minimum version to the v10 branch. It will force users to update their version of NodeJs, but it would allow us to use the `async` and `await` keywords.
@DeviaVir are you OK with that ?

The first commit update `travis.yml` to check everything is OK. I will add more commit if you're OK.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,19 @@
 language: node_js
 node_js:
   - '8'
-  - '9'
+  - '10'
+  - '11'
 before_install:
   - npm install -g node-gyp
-  - if [ "$TRAVIS_NODE_VERSION" = "9" ]; then npm install -g greenkeeper-lockfile@1; fi
+  - if [ "$TRAVIS_NODE_VERSION" = "10" ]; then npm install -g greenkeeper-lockfile@1; fi
 script:
   - npm run lint
   - npm test
 after_success:
-  - if [ "$TRAVIS_NODE_VERSION" = "9" ]; then greenkeeper-lockfile-update; fi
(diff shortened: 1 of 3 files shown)


DeviaVir:
Yes, I think supporting LTS and stable is enough. Definitely OK with this.

BAKFR:
Seeing the Travis results, we should fix tulind for Node v10 before anything.

I'm not even sure what's the problem is. When I try to install it with Node v10 on my side, it works.

BAKFR:
TulipCharts/tulipnode#48 should fix our install problems. It took me a while to understand this one.

Basically, a bug in node-pre-gyp cause the `tulind` build to be executed twice at the same time (for NodeJs >= 10).

BAKFR:
Travis tests still run with node-pre-gyp 0.10.3 instead of 0.12, and I don't know why :(

DeviaVir:
@BAKFR 
> Travis tests still run with node-pre-gyp 0.10.3 instead of 0.12, and I don't know why :(

forgot to update `package-lock.json`?

BAKFR:
> forgot to update `package-lock.json`?

That was it!
Now, everything is green

PR #1826, opened 2019-01-31
Fix order time on simulation

marian2js:
`now` is [used](https://github.com/DeviaVir/zenbot/blob/unstable/extensions/exchanges/sim/exchange.js#L113) but never set, this was causing the order time to be undefined on simulations


DeviaVir (approved):
good catch :+1:

PR #1837, opened 2019-02-24
Add missing period_length option to fix dema strategy

sdclibbery:
Just cloned Zenbot, and got a timebucket parse error when trying to run the dema strategy. Looks like the dema strategy hasn't been updated to use the `period_length` option in addition to the `period` option.


DeviaVir (approved):
great catch!

PR #1838, opened 2019-02-27
My contribs for zenbot!

marcelomf:
Last the 2 weeks, i worked in zenbot and try to be more profitable. Here are the changes i made. 
Currently I have been looking for strategies on top of the oderbook and on machine learning / HTF. I would like to reproduce this article: https://towardsdatascience.com/aifortrading-2edd6fac689d
I'm looking for help to make zenbot more profitable, or working on a new bot like https://github.com/jsappme/node-binance-trader or another.
Thanks.

### DIFF
--- a/extensions/strategies/vpt/strategy.js
+++ b/extensions/strategies/vpt/strategy.js
@@ -9,17 +9,13 @@ module.exports = {
     this.option('min_periods', 'min. number of history periods', Number, 50)
   },
 
-  // chamado primeiro, mas sem o lookback, trabalhar em s.period
+  // called first, but without lookback, working on s.period
   calculate: function(s) {
     if(s.lookback && s.lookback[0] && s.lookback[0].vpt){
       //s.period.vpt = s.lookback[0].volume + s.period.volume * ((s.period.close - s.lookback[0].close)/s.lookback[0].close)
       s.period.vpt = s.lookback[0].vpt + s.period.volume * ((s.period.close - s.lookback[0].close)/s.lookback[0].close)
     } else if(s.lookback && s.lookback[0]){
       s.period.vpt = s.period.volume + s.period.volume * ((s.period.close - s.lookback[0].close)/s.lookback[0].close)
-      // if(s.period.open > s.period.close)
-      //   s.trend = 'down'
-      // else
-      //   s.trend = 'up'
     }
   },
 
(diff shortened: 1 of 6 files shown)


extensions/strategies/vpt/strategy.js
```diff
@@ -0,0 +1,50 @@
...
+  description:
+    'Vpt - Volume Price Trend Indicator.',
+
+  getOptions: function () {
+    this.option('period', 'period length, same as --period_length', String, '30m')
+    this.option('period_length', 'period length, same as --period', String, '30m')
+    this.option('min_periods', 'min. number of history periods', Number, 50)
+  },
+
+  // chamado primeiro, mas sem o lookback, trabalhar em s.period
```
mrseeker: Can you translate this into English?

mrseeker:
commit 51bec76 needs some translations.

extensions/strategies/vpt/strategy.js
marcelomf: @mrseeker yes! Thanks.

BAKFR:
> Currently I have been looking for strategies on top of the oderbook and on machine learning / HTF

I'm not sure Zenbot can handle orderbooks at all.

> I'm looking for help to make zenbot more profitable, or working on a new bot like https://github.com/jsappme/node-binance-trader or another.

The more I work with Zenbot, the more I find bug that makes me think Zenbot never even remotely been correct. By example, I know before October 2018, the simulation results were completely broken (obvious errors in fee calculation). I've never been confident enough to use the sim mode.

To implement everything you need, either starting from scratch or from another code base could be a better choice.

marcelomf:
Today i did maked tests in coinbasepro(GDAX) with ETH-USD and the --reverse didn't make sense. But in binance did make be sense! I think we have bugs in binance related code.

rountrey:
Hi, I looked at your multi strategy and really liked it. I plan on using it after some modifications to use different strategies. I don't really code, so it might take some time to work it out. I do have a request, though. Could you put the totalBuy and totalSell into the onReport, please? I think it would be helpful to see those values change when it triggers a trade. Thanks.

DeviaVir (approved):
LGTM, thanks for contributing! :+1:

mmdiego:
This last commit breaks my custom strategy. As far as I could follow the code, it has to do with the change in engine.js, line 934 where you comment the line s.action = null.
The problem is that s.action remains in the state 'bought' for more than one period, and that causes some problems in my strategy, which is expecting s.action = bought to only last for one period and then go back to null.
If I revert that line, my strategy works well, although I get some messages about "buy cancel quarentine time...."
Can you please check the modification and propose a change to keep the old behaviour? Thanks!

Issue #1771, opened 2018-11-18
Problem with STDDEV strategy error


ghost:
Hello,
I'm using zenbot with this command line
./zenbot.sh trade --paper kraken.XXLM-ZEUR --strategy stddev --period=5000ms
it works for like 10 minutes, i can see some information about last exchange but it stops with this

/home/antoine/zenbot/extensions/strategies/stddev/strategy.js:24
      for (let i = 0; i < s.options.trendtrades_1; i++) { tl0.push(s.lookback[i].close) }
                                                                                 ^

TypeError: Cannot read property 'close' of undefined
    at Object.onPeriod (/home/antoine/zenbot/extensions/strategies/stddev/strategy.js:24:82)
    at onTrade (/home/antoine/zenbot/lib/engine.js:921:27)
    at /home/antoine/zenbot/lib/engine.js:899:5
    at /home/antoine/zenbot/node_modules/async/dist/async.js:4096:9
    at Object.process (/home/antoine/zenbot/node_modules/async/dist/async.js:2344:17)
    at /home/antoine/zenbot/node_modules/async/dist/async.js:2252:19
    at Immediate._onImmediate (/home/antoine/zenbot/node_modules/async/dist/async.js:119:16)
    at runCallback (timers.js:810:20)
    at tryOnImmediate (timers.js:768:5)
    at processImmediate [as _immediateCallback] (timers.js:745:5)

Can you help me because I would like to put this bot in a raspberry py in the futur, my actual system OS is kali linux

dcanones:
Same problem here, running Ubuntu 18.04 LTS in a VM:

```
(node:2836) DeprecationWarning: collection.save is deprecated. Use insertOne, insertMany, updateOne,│········································································································
 or updateMany instead.                                                                             │········································································································
2018-12-04 09:42:06  3491.19 BTC-EUR                 0                       0.01588 BTC  0.55 EUR  │········································································································
 -0.96%   -2.98%                                                                                    │········································································································
2018-12-04 09:42:25  3489.49 BTC-EUR   -0.05%        0                       0.01588 BTC  0.55 EUR  │········································································································
 -1.01%   -2.98%                                                                                    │········································································································
/home/david/repos/zenbot/extensions/strategies/stddev/strategy.js:24                                │········································································································
      for (let i = 0; i < s.options.trendtrades_1; i++) { tl0.push(s.lookback[i].close) }           │········································································································
                                                                                 ^                  │········································································································
[...]

PR #1843, opened 2019-03-17
Fix min_periods should be set after the strategy set its defaults #1771

lukavia:
If min_periods is set to 1 before the strategy had set its defaults actually makes the strategy defaults to be ignored and lead to some bugs like #1771 
To make mater worse the bug is reviewed only in trading. 
I think setting it last would fix the problem and not cause any problems


DeviaVir (approved):

DeviaVir:
Good find.

PR #1881, opened 2019-06-03
Added options --start and --end for backfill

durek1337:
The options start and end were not implemented but documented. In the documentation is the format for the option --start and --end this: YYYYMMDDhhmm. I took the unix-time in ms format like in the time property of the trades collection in the MongoDB.

### DIFF
--- a/commands/backfill.js
+++ b/commands/backfill.js
@@ -51,8 +51,8 @@ module.exports = function (program, conf) {
       }
       else {
         if(cmd.start >= 0 && cmd.end >= 0){
-          start_time = cmd.start;
-          target_time = cmd.end;
+          start_time = cmd.start
+          target_time = cmd.end
         } else {
           target_time = new Date().getTime()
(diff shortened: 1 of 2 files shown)


commands/backfill.js
```diff
@@ -48,8 +50,13 @@ module.exports = function (program, conf) {
         target_time = new Date().getTime() - (86400000 * cmd.days)
       }
       else {
-        target_time = new Date().getTime()
-        start_time = new Date().getTime() - (86400000 * cmd.days)
+        if(cmd.start >= 0 && cmd.end >= 0){
+          start_time = cmd.start;
+          target_time = cmd.end;
```
DeviaVir: can you remove the semicolons here?

durek1337: sure, it's done

DeviaVir: Cool, thanks!

PR #1892, opened 2019-06-14
hitbtc: update products

yellowgh0st:
- simply update Hitbtc.com exchange products


yellowgh0st:
I wonder if this will work because increment is now always 0.
Is this ok? I have no idea why is there.

DeviaVir:
hitbtc has an `update` script, which is automatically triggered by a cronjob that runs daily:
https://github.com/DeviaVir/zenbot/blob/unstable/scripts/cron.sh#L19-L25

Which means that your change would get overwritten tomorrow, unless it is generated by the update script.

PR #1903, opened 2019-07-04
add trades export to backtester_generation

felpasl:
improve analytics of trades and easily debug and make charts from estimated gains


DeviaVir (approved):
:+1: thanks for contributing

PR #1909, opened 2019-07-15
Fix some problems I encountered when trying to buy on bitstamp

r0bertz:


r0bertz:
I just found something interesting.

The following command works

```
$ zenbot buy bitstamp.XRP-USD  --debug

Bitstamp API is not answering! unable to call getQuote, retrying in 2s
got quote
not enough currency balance to buy!
```
'got quote' is a debug message that I added, not in this PR

But if I remove 'bitstamp.XRP-USD' (which is also the default selector in conf.js) from the command line , the command will enter an infinite loop

```
$ zenbot buy --debug

Bitstamp API is not answering! unable to call getQuote, retrying in 2s

Bitstamp API is not answering! unable to call getQuote, retrying in 2s

Bitstamp API is not answering! unable to call getQuote, retrying in 2s
placing order...

Bitstamp API is not answering! unable to call getQuote, retrying in 2s

Bitstamp API is not answering! unable to call getQuote, retrying in 2s
placing order...

```

r0bertz:
Ah ah. It looks like bitstamp/exchange.js gets currency pair from command line only. I will send another PR.

DeviaVir (approved):
LGTM. Look forward to your follow-up PR.

PR #1912, opened 2019-07-17
Error creating models file fixed.

bsesic:
While using the train command the model file is created with an attached '-simTrainingResult.html' which creates a .html file and crashes the app with an error, stating 'file not found'.
Removing the '-simTrainingResult.html' and the renaming lines solved the error.


DeviaVir (approved):
LGTM, thanks!

PR #1971, opened 2019-12-29
Fix spontaneous simulator exits

### LINKED ISSUE #1922: Zenbot suddenly stops during longer simulations
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: OSX
- **Zenbot version** (commit ref, or version): 4.1.0
- **Zenbot branch**: Master
- **NodeJS version**: v10.16.0
- **Exact command to reproduce (include everything)**: zenbot sim --strategy trend_ema --period_length 2m
- **Did I make any changes to conf-sample.js?**: `c.days = 300` and `c.currency_capital = 10000`

### Describe the problem

During longer simulations (~ more than 57 days for trend_ema) no matter the period, it suddenly stops without giving an output file or anything. It is different lengths depending on the strategy but none of them go ~ more than 100

### Source code / Error logs

--debug doesn't give any error logs not does anything else I can find, it just suddenly stops

jorisw:
Fixes #1922 by calling `getNext()` based on record count checks rather than the unreliable `.on('end')` cursor event.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -27,6 +27,7 @@
     },
     "parserOptions": {
         "sourceType": "module",
+        "ecmaVersion": "2017",
         "ecmaFeatures": {
             "experimentalObjectRestSpread": true
         }
--- a/commands/sim.js
+++ b/commands/sim.js
@@ -235,17 +235,18 @@ module.exports = function (program, conf) {
             process.exit(0)
           })
       }
-    
-      async function getNext () {
+
+      var getNext = async () => {
         var opts = {
           query: {
             selector: so.selector.normalized
           },
-          sort: {time: 1},
-          limit: 100
+          sort: { time: 1 },
+          limit: 100,
+          timeout: false
         }
         if (so.end) {
-          opts.query.time = {$lte: so.end}
+          opts.query.time = { $lte: so.end }
         }
         if (cursor) {
           if (reversing) {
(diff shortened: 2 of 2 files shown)


jorisw:
I see that ES6 style code isn't making it through eslint. If you want I can try and rewrite it into ES5.

DeviaVir:
> I see that ES6 style code isn't making it through eslint. If you want I can try and rewrite it into ES5.

@jorisw depends- does it make more sense to switch to ES6 at this point?

jorisw:
I could see that that may be a bit of a big move for zenbot, as indeed all of it has been written in ES5 so far. I personally never write ES5 anymore, and Node 8.3 (the minimum version for zenbot) runs all of my ES6 code without errors.

So ideally I suppose this PR needs to go to ES5, after which it _could_ be decided to take the project to ES6. However, as that would constitute a rewrite, that doesn't seem very feasible, therefore you could also simply decide to configure eslint to accept ES6 code, and accept this PR as it is.

jorisw:
I'm a bit puzzled actually as `es6` seems to be enabled in the project's `.eslintrc`. Does Travis use a config of its own?

jorisw:
Turns out I had `sim.js` in `.eslintignore` so that my IDE wouldn't reformat core files. I'll update the PR with something that validates.

jorisw:
I've ended up setting a modern ECMAScript version in `.eslintrc`, as my only alternative was to convert my changes to ugly Promise code. Since the project was already set to ES6, might as well allow developers to use await/async.

DeviaVir (approved):
:+1: great, thank you so much for contributing!

mmdiego:
I've updated my local copy of zenbot after some months and simulation stopped working after this commit. If I rolled back to the commit just before this one, which is ea7929f68635549e9546f5080e6050b9552b0d6a it starts working again.
I'm testing it using: 
`zenbot.js sim binance.BNB-USDT --strategy=macd --period=1m --days=1`
Maybe it's something related with the exchange?
I'm using node v8.3, but the same happens with v10.

Issue #1976, opened 2020-01-17
Simulation ends unexpectedly without generating the final report and html file.


brianunlam:
### System information
- **Have I written custom code (as opposed to using zenbot vanilla)**:no
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:Docker
- **Zenbot version** (commit ref, or version): 636bbf507df9944ab0311db04f826448ae5591a9
- **Zenbot branch**:unstable
- **NodeJS version**: v10.18.1
- **Python version (when using a python script)**:
- **Exact command to reproduce (include everything)**:zenbot sim binance.BTC-USDT --strategy rsi --days 1
- **Did I make any changes to conf-sample.js?**:
No
### Describe the problem
When running a simulation it ends unexpectedly when it reaches the last item from mongoDb. I think this is due to some changes applied to the logic to end the simulation a few weeks ago in the sim.js
### Source code / Error logs
I was able to find a fix. I can create a PR.

brianunlam:
Here I fixed the problem
https://github.com/DeviaVir/zenbot/pull/1977

PR #1977, opened 2020-01-17
Add check to end simulation and add drain function properly

brianunlam:
Simulation ends unexpectedly without generating the final report and html file.
This is because onCollectionCursorEnd() should be called in order to generate the final report. Checking when the amount of trades from mongoDb is zero, you can call that function to get the report. I have created an issue  #1976.


DeviaVir (approved):
:+1:

PR #1984, opened 2020-02-16
Zenbot install and Run with Raspberry

rasath:


with Raspberry
```diff
@@ -0,0 +1,56 @@
+**Zenbot Installation guide on Raspberry (Pi3 / Pi4)
```
DeviaVir, line 1: did you mean to update https://github.com/DeviaVir/zenbot/blob/unstable/docs/installation/raspberrypi.md ?

rasath: i have seen some docs about Zenbot can not run in Raspberry 1GB, but i follow this steps and its worked for me.
so i thought to share with this steps.

PR #1994, opened 2020-03-10
Add notifications to ADAMANT Messenger

adamant-al:
Notifications to decentralized blockchain messenger ADAMANT.

Params:
- `c.notifiers.adamant.fromPassphrase` is a passPhrase for sender's account like `cart angry pudding exhibit rule typical card image actress miracle ketchup shuffle`. Get new one on https://msg.adamant.im/. Be sure sender's account is not empty, get free tokens in the app.
- `c.notifiers.adamant.toAddresses` is an array of ADM addresses who will receive notifications. Example: `['U15744811633756100864', 'U7553468023140142965']`. They should be not empty also as their public keys should be known to the blockchain.
- `c.notifiers.adamant.nodes` is a list of ADAMANT nodes app may use to send notifications. Keep them default or run your own node.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,4 @@ gen.*.sh
 .env
 scripts/auto_backtester/backtesting_*.csv
 database/*
+.vscode/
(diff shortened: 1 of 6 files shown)


DeviaVir (approved):
:+1: looks good, thank you for contributing!

PR #2070, opened 2020-04-10
Notifier interaction: add events to notifiers, possibly fix #1550

### LINKED ISSUE #1550: [Feature Request] Notifier Interaction
### Describe the Feature
I would like the ability to "Talk" to zenbot through a connected notifier.  If, for example, I have discord configured with zenbot I would like to send the message "What are my current balances for Binance" and have it respond with all my current holdings.  Also have things like switching to manual mode so that it will stop trading.  There are many other examples as you can imagine.  But the first step is to create that framework.

emabo:
Implemented only for Telegram at the moment.
- list of accepted commands is the same used in trade.js
- check the correct origin of the message looking at chat ID
- accept flag 'c.notifiers.telegram.interactive' (default = false) to enable/disable this feature

This commit provides a possible solution for feature request #1550.


DeviaVir (approved):

DeviaVir:
Changes LGTM

Issue #2172, opened 2020-05-23
hitbtc doesn't submit trades


zaphbbrox:
## System information
- **Have I written custom code (as opposed to using zenbot vanilla)**:
  - No.
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:
  - Debian 9.12
- **Zenbot version** (commit ref, or version):
  - 4.1.0
- **Zenbot branch**:
  - unstable
- **NodeJS version**:
  - v14.3.0
- **Python version (when using a python script)**:
- **Exact command to reproduce (include everything)**:
  - `./zenbot.sh trade --paper hitbitc.BTC-USDT`
- **Did I make any changes to conf-sample.js?**:
  - only entered API key and XMPP credentials in **conf.js**

### Describe the problem
First of all, I'm aware there is a similiar open issue (#699), but since it wasn't updated in the last years I decided to open a new one.

After a successful installation everything seemed to work fine. `backfill`, `sim` and Web-UI are all working.
BUT:
In `trade` mode zenbot doesn't execute automatic orders, manual orders are only executed with `--paper` option. It's running now for several hours and in `sim` mode there were some trades in that time range.

### Source code / Error logs
Not sure what could be helpful, so just write me and I'll provide logs and stuff.
Debug mode gave me no output.

zaphbbrox:
A small update after scrolling trough the output of the last day:
- automatic orders are being executed, but still don't being pushed to hitbtc in live mode
- API can connect, since it's loading my balances correctly and I see a correct timestamp on hitbtc under "last use"
- also `timesyncd` is being used on the computer running zenbot

zaphbbrox:
After changing the trading pair to ETH-USDT it's working! Yay :grin: 
But still not sure why BTC trading is not working, since I was able to sell/buy on hitbtc, so a limit could not be in effect. Or are there different limits for the API use?
Nevertheless I close this issue. Feel free to re-open or ask anything.

Thanks for maintaining this repo! :star:

highlife22:
Hello, 
I have the same issue so i check,
in products.json for hitbtc, 

```javascript
 {
    "asset": "BTC",
    "currency": "USD",
    "min_size": 0.001,
    "increment": "0.00001",
    "label": "BTC/USD"
  }
```
Value from https://api.hitbtc.com/api/2/public/symbol/BTCUSD
```javascript
{
"id":"BTCUSD",
"baseCurrency":"BTC",
"quoteCurrency":"USD",
[...]

PR #2188, opened 2020-06-01
Update value for hitbtc BTCUSD

highlife22:
After i solve this issue #2172 

The value for BTC-USD (ex USDT) was not good or not good anymore

Min_size was too high and increment was to high
For some reason in Hitbtc the currency for BTC-USDT on the web site is different in the API it's BTC-USD

### DIFF
--- a/extensions/exchanges/hitbtc/products.json
+++ b/extensions/exchanges/hitbtc/products.json
@@ -8,10 +8,10 @@
   },
   {
     "asset": "BTC",
-    "currency": "USDT",
-    "min_size": 0.01,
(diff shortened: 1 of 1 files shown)


DeviaVir:
This update will be automatically rolled back by https://github.com/DeviaVir/zenbot/blob/184681917a6ba7ba3093581ff5001dc26db2ebdb/extensions/exchanges/hitbtc/update-products.sh

If you want to make sure the change is not rolled back, first run the `update-products.sh` locally and verify your change is still there. If it is rolled back, make changed to `update-products.sh` until it does not.

PR #2394, opened 2020-09-20
Tbw add binanceus

tbakerweb:
Adding Binance US Exchange


DeviaVir (approved):
:+1: Nice work. 

offtopic: I wonder if there's a more generic ccxt exchange we could write that would work with any of the exchanges ccxt supports.. but I'd guess they all have their own quirks.

tbakerweb:
This one was pretty straight forward.  I, ultimately, copied the binance exchange, and simply changed the name to binanceus, and added the required conf-sample content.

I’m very new to zenbot, only having found it this weekend.  I’m enjoying the project so far. I cut my teeth on trading bots with WolfBot, but faced far too many bugs.

Some of the features there have me thinking about investing some time into zenbot dev.  Most notably:

  *   Multi-Exchange support
  *   Multi-Currency Pair Support

I raise this, mostly, to determine if anyone else has work of this fashion going.  Any ideas?

PR #2555, opened 2020-12-12
Create howto_create_strategy.md

k3ntinhu:


DeviaVir (approved):

DeviaVir:
Great start, thanks for contributing this and making zenbot better!

the-aoji:
Hello,

Is there a command to query/store/recall previous order amounts? For example, I want to have zenbot retrieve the last 3 orders/trades that were filled, and the amount of asset/currency in that order. So if last order was for a sell for 10 ETH, and 5 ETH was sold, then the query would return that 5 ETH was the amount of the last order. 

Or if partially filled orders cannot be recalled, is there a command for completed order information to be queried? 

Thanks!

PR #2559, opened 2020-12-13
Add conf.js to .dockerignore and mount it as volume in docker-compose

sskender:
Add conf.js to .dockerignore to prevent rebuilding entire docker image when conf.js file is changed.
Instead, just mount it as volume to docker-compose.yml

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,6 +7,7 @@ services:
     image: fculpo/zenbot:latest
     volumes:
       - ./simulations/:/app/simulations/
+      - ./conf.js:/app/conf.js:ro
     restart: always
     tty: true
     ports:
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -7,6 +7,7 @@ services:
     image: fculpo/zenbot:latest
     volumes:
       - ./simulations/:/app/simulations/
+      - ./conf.js:/app/conf.js
```
DeviaVir: nit: Might want to mount it as read-only as to make sure we only change it from one side

e.g.

`- ./conf.js:/app/conf.js:ro`

sskender: Sure, great idea. I updated the pull request.

DeviaVir (approved):
:+1: great, thanks for contributing!

PR #2576, opened 2020-12-22
Helm chart/v1

Scizoo88:
Added a very basic implementation of a helm chart. There are not many configuration options at the moment for this chart. But it's a good starting point if you want to run ZenBot in a Kubernetes environment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,4 @@ gen.*.sh
 scripts/auto_backtester/backtesting_*.csv
 database/*
 .vscode/
+helm/zenbot/charts/*
--- a/helm/zenbot/.helmignore
+++ b/helm/zenbot/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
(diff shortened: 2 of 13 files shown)


xunholy:
@Scizoo88 thanks for putting in the effort to construct this helm chart, I took some time to give it a try and found a few issues with connecting to the MongoDB instance with auth and then an error due to a type error, may I ask if this was ever tested/working for you previously?

Scizoo88:
@xUnholy Hi, sure I run this helm chart on my RKE cluster at home. Did you put in the correct values in the values.yaml file?

xunholy:
What is your current installation method using your configuration?

I attempted to manually hydrate the manifests and apply them and i used the config.js that came with the latest version of zenbot.

Scizoo88:
@xUnholy I can post an detailed how-to on weekend. I see that in the values.yaml file, there is a  wrong Docker image configured, this was an alternative test from another repository and it seems that I didn't remove it, sorry. Can you test it with the official Image "deviavir/zenbot:latest"? I guess this is the problem why the config.js does not work.

djjudas21:
I was considering writing a Helm chart for Zenbot, so I'm pleased to see this. If I get some time this weekend I'll have a look at this and see if we can get it merge-able :+1:

PR #2582, opened 2020-12-27
Fix docker, remove public open MongoDB

XL-Reaper:
There are no dist files built, so they cannot be copied. This fixes the current docker issues.
Also I removed the public open MongoDB. It is enough if it is just reachable for zenbot through the docker networking.


stefancrain:
From my experience today : The removal of the copy commands causes the web portal at `:17365` to fail. Failing to copy over the `/app/dist` folder that is generated during the `RUN npm install *` layers causes multiple 404 messages when loading the webview. 
image

PR #2425, opened 2020-10-09
Fix: use lolex only for setTimeout function to avoid mongodb to stall

brianunlam:
Basically since version 3.6.2 mongodb library has replaced processWaitQueue with setImmediate.

lolex.install() is monkey patching setImmediate which is causing mongodb library to stall.

Here you can see more details about this:

https://github.com/mongodb/node-mongodb-native/pull/2537

https://github.com/Automattic/mongoose/issues/9417

So, in order to solve the problem we will only use lolex for setTimeout and not for other clock methods.

It fixes: https://github.com/DeviaVir/zenbot/issues/2412#issuecomment-705963100

PR #2600, opened 2021-01-05
Update engine.js

xwolfyxNL:
Reverts commit #2425 and fixes the broken sim results

### DIFF
--- a/lib/engine.js
+++ b/lib/engine.js
@@ -758,14 +758,7 @@ module.exports = function (s, conf) {
   }
 
   function withOnPeriod (trade, period_id, cb) {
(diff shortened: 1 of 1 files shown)


jorisw:
Could you elaborate, for future reference, why #2425 was bad and needed to be reversed?

xwolfyxNL:
> Could you elaborate, for future reference, why #2425 was bad and needed to be reversed?

It seems the faking of the data, causes the calculation of the loss/profit to increase by 10 fold. This causes profit margins of around 400% in a sim of a few days.

 I don't exactly why it does this, as I was just looking for a fix, but I backtracked it to commit #2425

PR #2615, opened 2021-01-17
trading with kraken

ChairmenMeow100:
I have encountered some problems while trading with kraken, so i updated and rearranged the  extensions/exchanges/kraken/products.json and altered the 42. line in extensions/exchanges/kraken/exchange.js to support all available coins

### DIFF
--- a/exchange.js
+++ b/exchange.js
@@ -39,7 +39,7 @@ module.exports = function container(conf) {
     var asset = product_id.split('-')[0]
     var currency = product_id.split('-')[1]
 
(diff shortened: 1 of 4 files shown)


DeviaVir:
Did you mean to place these files into https://github.com/DeviaVir/zenbot/tree/unstable/extensions/exchanges/kraken? They're currently in the root.

ChairmenMeow100:
exactly - in the meantime has kraken updated their pairs to support GBP and AUD so i have to re-update both files again ... you will get them soon.

ChairmenMeow100:
this should do the trick ... i am still unable to get CHF and DAI to work (maybe there is still a naming problem) but all of the other pairs work with backfill so i assume it works in general.

LuisAlejandro:
@ChairmenMeow100 The file is in the wrong place, you have to fix the pull request before it can be merged. The file goes in `extensions/exchanges/kraken/exchange.js`

ChairmenMeow100:
@LuisAlejandro sorry, i am completely new to all of this - now i see the problem ... is there a way to fix this pull request? I only found solutions that create a new pull request

PR #2648, opened 2021-02-18
remove wexnz

tommyyama2020:
remove an old exchange


DeviaVir:
Thanks for keeping zenbot up to date!

PR #2653, opened 2021-02-19
Added GDAX Sandbox option

Maddcow:
Added option to use Coinbase Pro sandbox  and created a gdax exchange doc covering the sandbox option and how to update fees.


DeviaVir (approved):
:+1: nice feature

PR #2657, opened 2021-02-23
Helm chart/v1

djjudas21:
This PR is an extension of the work in #2576 (thanks @Scizoo88)

This PR adds a Helm chart which deploys ZenBot and MongoDB with default settings. Instead of having to pass in a `conf.js` file, the individual settings are modelled as yaml data in `values.yaml` and can be overridden in the usual way for Helm charts.

I have tested this in MicroK8s - it successfully creates a MongoDB server and a ZenBot job running with dummy config. The job runs one-shot so when your trading period / sim finishes and ZenBot exits, Kubernetes does not restart the job!

To consider/discuss before merging:
* Hosting a Helm chart repo so this chart can be used easily


Scizoo88:
@djjudas21 Thanks for your updates to the chart.

djjudas21:
I'm running into a problem. Zenbot starts up normally but after some time, it crashes with this error:

```
/app/node_modules/mongodb/lib/utils.js:132
      throw err;
      ^

MongoError: limit requires an integer
    at Function.create (/app/node_modules/mongodb-core/lib/error.js:43:12)
    at Cursor.limit (/app/node_modules/mongodb/lib/cursor.js:636:22)
    at /app/commands/trade.js:426:66
    at result (/app/node_modules/mongodb/lib/utils.js:414:17)
    at executeCallback (/app/node_modules/mongodb/lib/utils.js:406:9)
    at handleCallback (/app/node_modules/mongodb/lib/utils.js:128:55)
    at cursor.close (/app/node_modules/mongodb/lib/operations/cursor_ops.js:211:62)
    at handleCallback (/app/node_modules/mongodb/lib/utils.js:128:55)
    at completeClose (/app/node_modules/mongodb/lib/cursor.js:893:14)
    at _endSession (/app/node_modules/mongodb/lib/cursor.js:904:37)
    at Cursor._endSession (/app/node_modules/mongodb-core/lib/cursor.js:194:5)
    at Cursor._endSession (/app/node_modules/mongodb/lib/cursor.js:226:59)
    at Cursor.close (/app/node_modules/mongodb/lib/cursor.js:904:19)
    at cursor._next (/app/node_modules/mongodb/lib/operations/cursor_ops.js:211:23)
    at handleCallback (/app/node_modules/mongodb-core/lib/cursor.js:203:5)
    at _setCursorNotifiedImpl (/app/node_modules/mongodb-core/lib/cursor.js:561:38)
    at self._endSession (/app/node_modules/mongodb-core/lib/cursor.js:569:46)
    at ClientSession.endSession (/app/node_modules/mongodb-core/lib/sessions.js:129:41)
    at Cursor._endSession (/app/node_modules/mongodb-core/lib/cursor.js:189:13)
    at Cursor._endSession (/app/node_modules/mongodb/lib/cursor.js:226:59)
    at _setCursorNotifiedImpl (/app/node_modules/mongodb-core/lib/cursor.js:569:17)
    at setCursorNotified (/app/node_modules/mongodb-core/lib/cursor.js:561:3)
```

I don't know anything about the Zenbot internals but is this something to do with passing in an integer config value as a string? Using Kubernetes to set environment variables casts everything as a string. If this is that problem, it probably means I can't carry on with this approach.

Scizoo88:
Have you tried to remove the "quote" in the deployment.yaml file?

djjudas21:
Yes, the `quote` is required because all values in an `env` hash must be strings. You have to quote integers and booleans, otherwise it throws an error.

Scizoo88:
Ok that's unfortunate. I guess the only other option to configure the conf.json via values.yaml would be to still bundle the conf.js with the chart but add as example:

`
c.mongo.db = process.env.ZENBOT_MONGODB_DATABASE || **{{ .Values.env.ZENBOT_MONGODB_DATABASE }}** || 'zenbot4'
`

Then in the Configmap use the "tpl" function:

`
{{ **tpl** (.Files.Get "resources/conf-sample.js") . | indent 4 }}
`

Remove the env declaration from the deployment.yaml file.

djjudas21:
Ah, I didn't know about the `tpl` function - thanks. I'll try this out a bit later :+1:

djjudas21:
I've pushed a lot more work into this. The two main changes are that the config is now managed as structured yaml which Helm uses to write out `conf.js`, and that the chart now runs ZenBot as a one-shot Job instead of a Deployment. This is so when ZenBot exits at the end of a sim etc, Kubernetes doesn't dive in and restart it. That could be very undesirable behaviour in a live environment where ZenBot has quit to prevent a loss, etc.

